PR #91, opened 2014-12-31
Update locales-tr-TR.xml

ekizyener:
I have surely done some changes in code for Turkish. Can you fix it? Thanks...

### DIFF
--- a/locales-tr-TR.xml
+++ b/locales-tr-TR.xml
@@ -62,10 +62,10 @@
     <term name="bc">M.Ö.</term>
 
     <!-- PUNCTUATION -->
-    <term name="open-quote">"</term>
-    <term name="close-quote">"</term>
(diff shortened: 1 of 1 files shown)


rmzelle:
@ekizyener, so the only issue is that you have trouble typing smart quotes? If so, you can just copy and paste them from https://github.com/citation-style-language/locales/blob/master/locales-en-US.xml#L65 (or from http://en.wikipedia.org/wiki/Quotation_mark#Summary_table_for_all_languages, or from MS Word).

ekizyener:
As you said i did changes on scripts.
Thanks.

adam3smith:
wonderful, thank you! This will be in the next Zotero release (IIRC that's your ref manager of choice).

rmzelle:
Thanks!

PR #92, opened 2015-01-27
Update locales-ru-RU.xml

myurkin:
Translating new English terms and implementing gender-specific ordinals.
Unfortunately, I am not aware of any way to automatically test the locale file, as it is done for styles.

### DIFF
--- a/locales-ru-RU.xml
+++ b/locales-ru-RU.xml
@@ -26,10 +26,10 @@
     <term name="anonymous">аноним</term>
     <term name="anonymous" form="short">анон.</term>
     <term name="at">на</term>
-    <term name="available at">available at</term>
+    <term name="available at">доступно на</term>
     <term name="by"/>
-    <term name="circa">circa</term>
-    <term name="circa" form="short">ca.</term>
+    <term name="circa">около</term>
+    <term name="circa" form="short">ок.</term>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!
http://validator.citationstyles.org/
works for style files, too. See if that helps you figure out the problems here, otherwise we'll take a closer look later.

rmzelle:
See http://citationstyles.org/downloads/specification.html#gender-specific-ordinals and http://validator.citationstyles.org/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fmyurkin%2Flocales%2Fmaster%2Flocales-ru-RU.xml&version=1.0.1 .

The problem is that you defined the gender for terms that don't have a matching "number variable".

myurkin:
I have fixed the errors. 
As a side note for future development - it seems logical to make all locator terms gender-assignable, since ordinal numbers may potentially be used in locators (which are manually entered in Word), e.g. Authors, Book, Publisher, (year). (2nd chapter, 37th page).

adam3smith:
thanks! Makes sense re: locators. @rmzelle -- am I missing something?

rmzelle:
Can't find a fault in the logic in that.

PR #95, opened 2015-02-02
Translated English terms to (Swiss) German

tillheilmann:


locales-de-CH.xml
```diff
@@ -199,8 +199,8 @@
 
     <!-- LONG ROLE FORMS -->
     <term name="director">
-      <single>director</single>
-      <multiple>directors</multiple>
+      <single>Regisseure</single>
```
zuphilip: typo?

PR #103, opened 2015-02-19
Inserted non-breaking spaces

tillheilmann:
... to prevent line breaks in abbreviations and before ampersands.


adam3smith:
and thanks again.

PR #106, opened 2015-06-09
Create locales-es-MX.xml

natch-nacht:
For use in Mexico.


rmzelle:
Does Mexican Spanish deviate from the other Spanish locales we have?

natch-nacht:
I relied on the xml on locales-es.CL.mxl, because there are slight linguistic differences that preclude me to use Chilean xml file. 

I don't know if I used the <link href="https://github.com/citation-style-language/locales/blob/master/locales-es-CL.xml" rel="templete"/> line correctly on <info> section.

PR #107, opened 2015-06-09
locales-es-MX

natch-nacht:

PR #108, opened 2015-06-28
Create locales-es-MX.xml

natch-nacht:
For use in Mexico.

### DIFF
--- a/locales-es-MX.xml
+++ b/locales-es-MX.xml
@@ -4,7 +4,7 @@
     <translator>
       <name>Juan Ignacio Flores Salgado</name>
       <uri>https://www.mendeley.com/profiles/juan-ignacio-flores-salgado/</uri>    
-	  </translator>
+    </translator>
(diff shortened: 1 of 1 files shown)


rmzelle:
@adam3smith, can you take a brief look? Looks good to me: https://www.diffchecker.com/r2go2zet

adam3smith:
thanks!

rmzelle:
(and this continues from #106 and #107)

rmzelle:
(oh, and @adam3smith, we should make sure to update https://github.com/citation-style-language/locales/blob/master/locales.json and https://github.com/citation-style-language/locales/wiki when accepting new locales (I'll do that now); and verify that the language tag is correct, which it seems to be per http://r12a.github.io/apps/subtags/)

rmzelle:
Added to locales.json: https://github.com/citation-style-language/locales/commit/17bc27f4a1835b9f23e5daea98fb61ee2af94072

PR #111, opened 2015-07-25
Fix month to numeric in numeric date form in fi-FI

jhuovari:
In fi-Fi the date form="numeric" had month in a text form. This will fix it to numeric. The Finnish date format is: d.m.yyyy.


rmzelle:
Great, thanks! That mistake has been there a while.

PR #112, opened 2015-08-05
Create locales-id-ID.xml

xbypass:
Indonesia is the fourth most populous nation in the world. Of its large population, the majority speak Indonesian, making it one of the most widely spoken languages in the world. Thus, it is necessary to give this large population locales.


adam3smith:
couldn't agree more. I always wondered why no one complained about this. Thanks for doing this!

rmzelle:
Thanks!

PR #113, opened 2015-08-28
Corrected quotation marks in vi-VN locale.

dowens76:
Perhaps in previous generations Vietnamese writers used French quotes
or guillemets, but I am not aware of a publisher in Vietnam today that
uses them. They use quotation marks as in English. As evidence, see a
famous newspaper in Vietnam (http://tuoitre.vn/). If necessary, I can
provide documentation from book publishing.


adam3smith:
yeah, we based this on Wikipedia without any actual knowledge of Vietnamese, so I'm sure you're right. Thanks also for the other additions.

Issue #115, opened 2015-09-16
Use fol./fols(.) for short form of 'folio' term?


adunning:
CMoS (14.162, 14.235) and MHRA both give 'fol.', plural 'fols' (with period in en-US, often omitted in en-GB) rather than the CSL 'f.', which occurs but is unusual. Should this be changed?

adam3smith:
in general yes -- quesiton is where. I'd say if this is both Chicago and MHRA, we should just change it in the locale. 
We could consider doing the plural w/o period in en-GB. @rmzelle what do you think?

rmzelle:
Do any of the current styles redefine this term? I agree with "fol/fols". Just an "f" seems ambiguous.

adam3smith:
yes, a couple:
grep -l term\ name=\"folio\" *.csl
chicago-author-date-basque.csl
irish-historical-studies.csl
jahrbuch-der-osterreichischen-byzantinischen-gesellschaft.csl
journal-of-neurochemistry.csl
medical-history.csl
modern-humanities-research-association-author-date.csl
modern-humanities-research-association.csl
past-and-present.csl
u-schylku-starozytnosci.csl

some of these probably don't define the English one, though.

adam3smith:
Also, what do you think about the period? Should we do it w/o period in the plural in en-GB?

adunning:
The rest of the en-GB file uses periods (e.g. 'vols.'), but it's not clear to me whether there is a technical reason for this, or whether it's an oversight. The most common practice in British usage (see e.g. MHRA, New Hart’s Rules) would write 'fols', 'vols', 'cols', 'eds', etc. without the period, the rationale being that these are all contractions: the exceptions are 'no.' and 'nos.' for the sake of clarity.

rmzelle:
The reason we prefer to have periods in locale file terms is that removing periods is much easier to automate than adding them. (since the computer doesn't know which terms are abbreviations, and which aren't)

Leaving out the periods is only a good idea if you're positive that a term (almost) never uses periods.

adam3smith:
the problem here, though, is that only the plural term doesn't, so strip-periods doesn't work. I've usually redefined these in the style, i.e.
```
<term name="editor" form="short">
  <single>ed.</single>
  <multiple>eds</multiple>
</term>
```
My problem with redefining this would be that we'd have to check all existing en-GB styles because while this is common, it is not universal. Might still be worth it, but there are 240 of them...

rmzelle:
[...]

PR #116, opened 2015-09-21
Add plurals for all short locator forms in English.

adunning:
This adds plurals for all locator forms. The en-US list follows the Chicago Manual of Style, which does not specify all plurals: 'paras.' and 'secs.' are somewhat debatable.

I have looked at MHRA and New Hart's Rules for en-GB. Neither of these include an abbreviation list, so I have not changed any of the base definitions, though I think 'ch.' for chapter and 'par.' for paragraph is more common (and these are also alternatives in CMoS).

Some of the short verb role forms do not follow normal en-GB usage, but I am leaving this alone for now, per the discussion in #115.

### DIFF
--- a/locales-en-GB.xml
+++ b/locales-en-GB.xml
@@ -2,7 +2,7 @@
 <locale xmlns="http://purl.org/net/xbiblio/csl" version="1.0" xml:lang="en-GB">
   <info>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
-    <updated>2012-07-04T23:31:02+00:00</updated>
+    <updated>2015-09-21T23:31:02+00:00</updated>
   </info>
   <style-options punctuation-in-quote="false"/>
   <date form="text">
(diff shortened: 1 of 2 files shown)


adunning:
Just to check, is there anything further I should do with this?

adam3smith:
no, you're good. I'll get to it over the weekend.

adunning:
Great; thanks!

rmzelle:
I tried to find a few more style guides, and I searched Google for "cite apa paras" (without quotes). I found:

* http://www.slideshare.net/COTRreference/canadian-legal-decisions-in-apa-citation-style, Canadian, uses "paras"
* http://www.oxfordjournals.org/our_journals/jicjus/for_authors/stylesheet%202009.pdf, British, says "The Journal follows the style Art./Arts, ed./eds, para./paras, vol./vols i.e. no full stop after the abbreviated plural."
* http://libguides.macewan.ca/WLS/APA/sources-with-no-page-numbers, Canadian, uses "paras."
* http://www.informationr.net/ir/author2.html, says submissions should use British English, uses "paras."

So it's a bit hit-and-miss, and it looks like quite a few en-GB journals use the periods for the plural version. MHRA and Hart's Rules are probably more authoritative here, though, so this PR might still be the best way to go.

(also, it would be preferable if all locale files were extended with singular and plural forms of these short locator forms. Unfortunately that's easier said than done. I once regenerated all locale files by taking the en-US locale, and for each locale file in our repository replacing all en-US terms by the terms from that locale file. That ensured that all locale files had the same set of CSL terms and had the same formatting. Might be overkill for this change, but we might want to repeat this process at some later point just to keep things tidy.)

adunning:
Canadian usage is entirely mixed on this point. New Hart's Rules actually notes that some styles use 'vols.' (etc.) with a period, but their examples use it without.

Before you regenerate everything, take a look to check whether you want to fill out the list of verb role forms; I've been wondering whether the complete list should be provided in both the long and short versions.

adunning:
Just to check, does anything further need to be done with this? (Would like to get it into the next release of `pandoc-citeproc`.)

adunning:
Thank you!

rmzelle:
Thanks @adunning!

PR #117, opened 2015-09-23
Create el-Gr_locale_corrections.csl

thanasis57:
-edition (έκδοση) defined as feminine
-volume (τόμος) defined as masculine
-translated reviewed-author term. This is a bit tricky because we need a translation that is correct to use with the nominative declension of the reviewed author's name, which is the way authors are normally entered in the database. E.g., a translation of "by" with "του/της" would require the genitive form of the author's name, which would mean that an entirely new field would need to be created, i.e. the reviewed author's name in the genitive form.

### DIFF
--- a/el-Gr_locale_corrections.csl
+++ b/el-Gr_locale_corrections.csl
@@ -1,0 +1,308 @@
+<?xml version="1.0" encoding="utf-8"?>
+<locale xmlns="http://purl.org/net/xbiblio/csl" version="1.0" xml:lang="el-GR">
+  <info>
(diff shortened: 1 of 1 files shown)


gracile-fr:
I think you should edit the existing file https://github.com/citation-style-language/locales/blob/master/locales-el-GR.xml rather than creating a new file (```el-Gr_locale_corrections.csl``` in your case).

thanasis57:
You are probably right, I just followed the instructions for a pull request.

I will do so.

adam3smith:
superseded by https://github.com/citation-style-language/locales/pull/119
Thanks @gracile-fr for helping out.

PR #118, opened 2015-09-23
locales-el-GR-translations.xml

thanasis57:
-edition (έκδοση) defined as feminine, volume (τόμος) defined as masculine, and some others
-translated reviewed-author term. This is a bit tricky because we need a translation that is correct to use with the nominative declension of the reviewed author's name, which is the way authors are normally entered in the database. E.g., a translation of "by" with "του/της" would require the genitive form of the author's name, which would mean that an entirely new field would need to be created, i.e. the reviewed author's name in the genitive form.

### DIFF
--- a/locales-el-GR.xml
+++ b/locales-el-GR.xml
@@ -27,7 +27,7 @@
     <term name="circa">περίπου</term>
     <term name="circa" form="short">περ.</term>
     <term name="cited">παρατίθεται</term>
-    <term name="edition">
+    <term name="edition"  gender="feminine">
       <single>έκδοση</single>
       <multiple>εκδόσεις</multiple>
     </term>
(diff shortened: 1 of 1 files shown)


adam3smith:
To explain why this is failing: you can't assign a gender to those terms because CSL doesn't apply them in gender-sensitive contexts. Different for edition, which is why it works there.

thanasis57:
Thanks for the clarification. After I deleted them it was successful in pull 119.

rmzelle:
@thanasis57, for the future, it's generally preferred to update the original pull request instead of creating a new one if the original one has a problem. We can tell you how next time.

rmzelle:
(and this PR is superseded by #119)

PR #119, opened 2015-09-23
locales-el-GR.xml

thanasis57:
-edition (έκδοση) defined as feminine
-translated reviewed-author term and corrected illustrator name. This is a bit tricky because we need a translation that is correct to use with the nominative case of the reviewed author's name, which is the way authors are normally entered in the database. E.g., a translation of "by" with "του/της" would require the genitive form of the author's name, which would mean that an entirely new field would need to be created, i.e. the reviewed author's name in the genitive form. The workaround for Greek is to use a colon.

### DIFF
--- a/locales-el-GR.xml
+++ b/locales-el-GR.xml
@@ -27,7 +27,7 @@
     <term name="circa">περίπου</term>
     <term name="circa" form="short">περ.</term>
     <term name="cited">παρατίθεται</term>
-    <term name="edition">
+    <term name="edition" gender="feminine">
       <single>έκδοση</single>
       <multiple>εκδόσεις</multiple>
     </term>
@@ -69,7 +69,9 @@
     <term name="page-range-delimiter">–</term>
 
     <!-- ORDINALS -->
-    <term name="ordinal">ος</term>
+    <term name="ordinal">ο</term>
+    <term name="ordinal-01" gender-form="feminine" match="whole-number">η</term>
+    <term name="ordinal-01" gender-form="masculine" match="whole-number">ος</term>
 
     <!-- LONG ORDINALS -->
     <term name="long-ordinal-01">πρώτος</term>
@@ -253,10 +255,10 @@
     <term name="director" form="verb">διεύθυνση</term>
     <term name="editor" form="verb">επιμέλεια</term>
     <term name="editorial-director" form="verb">διεύθυνση σειράς</term>
-    <term name="illustrator" form="verb">εικονογράφηση by</term>
+    <term name="illustrator" form="verb">εικονογράφηση:</term>
     <term name="interviewer" form="verb">συνέντευξη</term>
     <term name="recipient" form="verb">παραλήπτης</term>
-    <term name="reviewed-author" form="verb">by</term>
+    <term name="reviewed-author" form="verb">συγγραφέας:</term>
     <term name="translator" form="verb">μετάφραση</term>
     <term name="editortranslator" form="verb">μετάφραση και επιμέλεια</term>
 


adam3smith:
thanks!

rmzelle:
@adam3smith, I don't think we should have punctuation like colons in term translations:

`<term name="illustrator" form="verb">εικονογράφηση:</term>` and `<term name="reviewed-author" form="verb">συγγραφέας:</term>`

adam3smith:
@rmzelle I found the reasoning compelling: 
> E.g., a translation of "by" with "του/της" would require the genitive form of the author's name, which would mean that an entirely new field would need to be created, i.e. the reviewed author's name in the genitive form. The workaround for Greek is to use a colon.

Don't know Greek, obviously, but I wouldn't know what else we can do.

rmzelle:
Ah, right. I'm just concerned about styles that need "illus by: John Doe" in English. With `<term name="illustrator" form="verb-short">illus. by</term>` and the Greek with-colon translation, that would result to a double colon if the style localizes to Greek.

thanasis57:
I could propose a couple of additional workarounds, if it so happens that such a conflict is reported for a particular style in the future. However, they probably would not look as good, be less elegant, or might be technically more complicated to implement:
-use "του/της". This would require to assign a gender attribute to the reviewed author and instruct users to always enter the reviewed author's name in the genitive case
-use "υπό". It is more archaic, but at least does away with the need to attribute a gender. The reviewed author's name still has to be in genitive case.
-use the form "(συγγρ. Firstname Lastname)", which would require to introduce an opening and a closing bracket, i.e. create a new style.

Still, the issue you raise reinforces my argument (https://forums.zotero.org/discussion/52518/corrections-for-elgr-greek-citation-styles/#Comment_234722) that it would make sense to have language-specific versions of styles, at least for certain languages that are quite different from English (in terms of alphabet, punctuation, grammar etc), whose intricacies are not covered by locales alone.

So, in case such a conflict arises, the interested author/editor/publisher could have the option to propose a localised version of a specific style (following convincing justification for doing so, of course).

rmzelle:
(I never knew that there are languages that have different cases for names of individuals; E.g. Latin doesn't do that, right?)

The drawback of language-specific versions of styles is that it becomes more difficult for us to keep all of them up-to-date. But you're right that in some cases CSL's automatic localization doesn't suffice, and CSL is indeed developed from a mostly English-centric perspective, although most CSL developers are multilingual. We don't have anybody who speaks Greek, though :).

adam3smith:
> (I never knew that there are languages that have different cases for names of individuals; E.g. Latin doesn't do that, right?)

yes, Latin does that, too. E.g. Principia philosophiae cartesianae by Benedictus of Spinoza are "per Benedictum de Spinoza" i.e. accusative case.

rmzelle:
Guess I should have paid more attention in high school!

thanasis57:
I don't know if Latin does that with all names, but it has cases for names inherited from Greek:

nomin. Penelope                      Aeneas/Aenea
gen.    Penelopae/Penelopes    Aeneae 
dat.     Penelopae                    Aeneae 
accus. Penelopes                     Aenean/Aeneam 
acc.    Penelopen                     Aenea/Aeneas
voc.    Penelope                      Aenea

And don't beat yourself up, i just looked it up myself!

https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=9&ved=0CF4QFjAIahUKEwjU15C_sJDIAhVMLB4KHTvdCLQ&url=http%3A%2F%2Fevgeniageorgakopoulou.weebly.com%2Fuploads%2F3%2F1%2F4%2F8%2F31483521%2Fklisi_oysiastikwn.pdf&usg=AFQjCNHJ5JVzqeEEHZ5c39-c3OcKliTbYw&bvm=bv.103388427,d.dmo

gracile-fr:
<blockquote> that would result to a double colon if the style localizes to Greek.</blockquote>Would there be any drawbacks to suppress this kind of double punctuation (I assume the different CSL implementations could do that, no)? I don't think that double colon can be a desired behaviour, but localization is full of surprises.

rmzelle:
> Would there be any drawbacks to suppress this kind of double punctuation

Even if we can fix it in the CSL processor, I'm more concerned about keeping the locale files tidy. But the colon might be the best solution here.

PR #121, opened 2015-10-11
Add short reviewed-author verb role form to English, French.

adunning:
There are over a dozen styles that call for this, which causes problems
for pandoc-citeproc; it is more straightforward to add it here than to
track down each occurrence.

### DIFF
--- a/locales-en-GB.xml
+++ b/locales-en-GB.xml
@@ -2,7 +2,7 @@
 <locale xmlns="http://purl.org/net/xbiblio/csl" version="1.0" xml:lang="en-GB">
   <info>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
-    <updated>2015-10-10T23:31:02+00:00</updated>
+    <updated>2015-10-11T23:31:02+00:00</updated>
   </info>
   <style-options punctuation-in-quote="false"/>
   <date form="text">
@@ -308,6 +308,7 @@
     <term name="editor" form="verb-short">ed. by</term>
     <term name="editorial-director" form="verb-short">ed. by</term>
     <term name="illustrator" form="verb-short">illus. by</term>
+    <term name="reviewed-author" form="verb-short">by</term>
     <term name="translator" form="verb-short">trans. by</term>
     <term name="editortranslator" form="verb-short">ed. &amp; trans. by</term>
 
--- a/locales-en-US.xml
+++ b/locales-en-US.xml
@@ -2,7 +2,7 @@
 <locale xmlns="http://purl.org/net/xbiblio/csl" version="1.0" xml:lang="en-US">
   <info>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
-    <updated>2015-10-10T23:31:02+00:00</updated>
+    <updated>2015-10-11T23:31:02+00:00</updated>
   </info>
   <style-options punctuation-in-quote="true"/>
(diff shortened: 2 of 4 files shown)


adunning:
(I tried to correct the error in the styles themselves, but it's slightly complicated in places such as <https://github.com/citation-style-language/styles/blob/be48829d31f51c9bf52da8e59a6df7ca706a7b3b/chicago-annotated-bibliography.csl#L56>.)

adam3smith:
I don't have a problem with this, and unless @rmzelle objects, I'll merge it tomorrow, but pandoc-citeproc should really handle this more elegantly: for variables, we explicitly state that "long" is used where "short" is missing. We should probably make this more explicit in the specs, but the same should happen with terms.

adunning:
It's certainly incorrect behaviour (<https://github.com/jgm/pandoc-citeproc/issues/72>), but on the other hand it is odd to have styles asking for a nonexistent form. Whatever you think will be the most robust solution.

adam3smith:
I'll see what Rintze says. I think implementing the fallback would be more robust for sure, but on the other hand this won't hurt.

rmzelle:
> for variables, we explicitly state that "long" is used where "short" is missing. We should probably make this more explicit in the specs, but the same should happen with terms. 

The CSL 1.0.1 spec is already very explicit about requiring term fallback behavior:

> If a style uses a term in a form that is undefined (even after Locale Fallback), there is fallback to other forms: “verb-short” first falls back to “verb”, “symbol” first falls back to “short”, and “verb” and “short” both fall back to “long”. If no locale or form fallback is available, the term is rendered as a empty string.

(http://docs.citationstyles.org/en/stable/specification.html#terms)

Also see the related

> Locale fallback takes precedence over fallback of term forms

(http://docs.citationstyles.org/en/stable/specification.html#locale-fallback)

Not adding redundant term forms reduces the work load for translators and keeps the files smaller. There is also a potential problem that could pop up if we only add this term to a select few locale files. Say we merge this pull request, and a CMoS CSL style is used with the Dutch locale file. If the style calls for the "verb-short" term, it will use the "en-US" translation, since locale fallback takes precedence over term form fallback. So we should be careful here.

adam3smith:
ah yes, that's why I waited for your comments (I looked in the wrong place for this in the specs); hadn't thought about the cross-language fallbacks This makes perfect sense to me, and given that I think the case for not merging is stronger.
I'd ask jgm to fix this in pandoc-citeproc.

adunning:
Thanks for looking at this! (I reported it over a year ago, and was hoping to find some other way out.) Given this, I certainly agree.

Is it an error that `container-author` is defined for `verb-short` but not for `verb`?

rmzelle:
> Is it an error that container-author is defined for verb-short but not for verb?

I did some sleuthing, and this goes way back to 2010:

https://github.com/citation-style-language/locales/commit/8b2de062d3fd128ca1ec58e0137285acddb2dfb5
http://xbiblio-devel.2463403.n2.nabble.com/container-author-td4629177.html

I think there was only a request for "verb-short", so we never added "verb". Not an error, but also no real reason not to add it.

adunning:
Thanks for that. It's no longer an issue for me (the pandoc-citeproc bug is fixed in 0.8), but I'm wondering whether `container-author` should be redefined as `verb`, meaning that any styles referring to this will fall back, but it would still work if someone accidentally referred to it without the `-short`. (Unless I am missing some other rationale.)

adam3smith:
makes sense to me; @rmzelle ?

rmzelle:
That would be better, yes. @adunning, are you up for preparing a PR to change this in all locale files (and move the "verb" translation to its proper spot in each locale file)? If you don't know how to automate this, I can probably figure out how to do it with a bit of RegEx.

PR #122, opened 2015-10-15
Move container-author from verb-short to verb.

adunning:
As discussed in #121, styles that refer to the term will continue to use this as a fallback. Also replaced empty terms with self-closing tags where this was not already done.

Note that `zh-TW` also includes an `author` term; I believe this is an oversight, but am leaving it there for now.

### DIFF
--- a/locales-cs-CZ.xml
+++ b/locales-cs-CZ.xml
@@ -251,6 +251,7 @@
     </term>
 
     <!-- VERB ROLE FORMS -->
+    <term name="container-author" form="verb"/>
     <term name="director" form="verb">řídil</term>
     <term name="editor" form="verb">editoval</term>
     <term name="editorial-director" form="verb">editoval</term>
@@ -262,7 +263,6 @@
     <term name="editortranslator" form="verb">editoval a přeložil</term>
 
     <!-- SHORT VERB ROLE FORMS -->
-    <term name="container-author" form="verb-short"></term>
     <term name="director" form="verb-short">řed.</term>
     <term name="editor" form="verb-short">ed.</term>
     <term name="editorial-director" form="verb-short">ed.</term>
--- a/locales-cs-CZ.xml
+++ b/locales-cs-CZ.xml
@@ -251,7 +251,7 @@
     </term>
 
     <!-- VERB ROLE FORMS -->
-    <term name="container-author" form="verb"/>
+    <term name="container-author" form="verb"></term>
     <term name="director" form="verb">řídil</term>
     <term name="editor" form="verb">editoval</term>
     <term name="editorial-director" form="verb">editoval</term>
--- a/locales-km-KH.xml
+++ b/locales-km-KH.xml
@@ -256,6 +256,7 @@
     </term>
 
     <!-- VERB ROLE FORMS -->
+    <term name="container-author" form="verb">by</term>
     <term name="director" form="verb">directed by</term>
     <term name="editor" form="verb">edited by</term>
     <term name="editorial-director" form="verb">edited by</term>
@@ -267,7 +268,6 @@
     <term name="editortranslator" form="verb">edited &amp; translated by</term>
 
     <!-- SHORT VERB ROLE FORMS -->
-    <term name="container-author" form="verb-short">by</term>
     <term name="director" form="verb-short">dir.</term>
     <term name="editor" form="verb-short">ed.</term>
     <term name="editorial-director" form="verb-short">ed.</term>
--- a/locales-km-KH.xml
+++ b/locales-km-KH.xml
@@ -213,7 +213,7 @@
       <multiple>editors</multiple>
     </term>
     <term name="editorial-director">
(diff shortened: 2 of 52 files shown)


locales-cs-CZ.xml
```diff
@@ -251,6 +251,7 @@
     </term>
 
     <!-- VERB ROLE FORMS -->
+    <term name="container-author" form="verb"/>
```
rmzelle: Are we sure that all XML parsers treat `<term></term>` and `<term/>` the same?

rmzelle: (cc @adunning, @adam3smith)

fbennett: What is the desired behaviour with singletons? Should the processor stop fallback and return an empty string, or treat the node value as nil? 

If the latter, and this is just a placeholder, why not make it a comment instead?

rmzelle: http://docs.citationstyles.org/en/stable/specification.html#locale-fallback says

> Fallback stops once a localizable unit has been found. For terms, this even is the case when they are defined as empty strings (e.g. `<term name="and"/>` or `<term name="and"></term>`).

adunning: Good to know; but I wonder if that's always the intent of the localizers. (Sometimes the terms are actually removed, e.g. `zn-CN` and `zh-TW` with most of the ordinals.)

fbennett: I guess that must be what we do then. To the question, I have seen
differential treatment of singletons in HTML parsers (some will drop the
node), but XML should (should) be consistent. I'll check what citeproc-js
does with it later, fwiw.

rmzelle: Well, ordinals are a whole other story (at least for the "ordinal-xx" terms). (see http://docs.citationstyles.org/en/stable/specification.html#ordinal-suffixes )

rmzelle: Anyway, the point I was trying to make is that it would be safer to keep the `<term></term>` format, just in case parsing is different in some of the CSL processors, although I'm not sure this is a real concern.

adunning: The files are already doing this in several places (see e.g. `ar` above), so it should have come out already if they are being handled differently.

rmzelle: Oh, and I just remembered that I prefer the `<term></term>` format because it's clearer to translations that we're dealing with an empty string. It's also easier for translations to turn `<term></term>` into `<term>blah</term>` than it is to turn `<term/>` into `<term>blah</term>` (the latter requires more editing and some understanding of HTML/XML tags).

It might be easier to fix this up for the entire repo, though, with a find-and-replace (basically "/>" to "></term>", although that would certainly mess up `<style-options/>` so some fancier RegEx would be needed). I can just merge this as is.

adunning: Yes, good point. Whatever you like; I can adjust it if desired.

rmzelle: If you don't mind?

locales-km-KH.xml
```diff
@@ -213,7 +213,7 @@
       <multiple>editors</multiple>
     </term>
     <term name="editorial-director">
-      <single/>
+      <single></single>
```
adunning: This is odd, but I am leaving it as-is for now in case it was removed for a reason.

locales-zh-TW.xml
```diff
@@ -148,6 +148,7 @@
 
     <!-- VERB ROLE FORMS -->
     <term name="author" form="verb">著</term>
```
adunning: Is this actually used in any style?

rmzelle:
Awesome, thanks.

adunning:
You're very welcome.

locales-zh-TW.xml
rmzelle: No clue. I would assume we would hear an outcry from Japanese users, though, since we use this term a lot in "author" macros that have substitution with the "editor" and/or "translator" variables (in which case the term is usually desirable).

adunning: Good to know; just wanted to make sure that wasn't a typo for `container-author`.

locales-km-KH.xml
fbennett: Khmer was contributed by a former student who does ref manager training in Phnom Penh. I'll ping him so he can take a look. Meanwhile as-is is fine.

locales-zh-TW.xml
adam3smith: Taiwanese users that is. For Japanese we'd have someone we could ask...

fbennett: I can ask a student later today.

adam3smith: we *do* have someone! Thanks Frank

fbennett: According to today's straw poll, the term is not applied to author names in citations in TW, nor in CN.

PR #124, opened 2015-11-25
Update locales-sv-SE.xml

torfeur:
Changes shorts for swedish months. May, June and July are normally not abbreviated, equivalent to the german de-DE locale.


rmzelle:
Looks good. Thanks!

PR #126, opened 2015-12-21
Add some missing translations

georgd:


rmzelle:
Could you move all instances of `collection-editor` up a little? The original listings are alphabetical, and I prefer to keep it that way.

rmzelle:
(you should be able to use the edit/pencil icon at https://github.com/georgd/locales/blob/patch-1/locales-de-AT.xml )

georgd:
I can do so, but I thought I’d adapt it to the [de-DE file] (https://github.com/citation-style-language/locales/blob/master/locales-de-DE.xml which shows the same order. Also, this order seems logical to me in analogy to the order of editor, translator and editortranslator in the original listing (en-US, I believe, is the original listing).

rmzelle:
Okay, sure, let's keep things as they are, then. Thanks!

PR #127, opened 2016-01-03
Update locales-id-ID.xml

xbypass:
Replace "by" with "oleh".

### DIFF
--- a/locales-id-ID.xml
+++ b/locales-id-ID.xml
@@ -23,7 +23,7 @@
     <term name="anonymous" form="short">anon.</term>
     <term name="at">pada</term>
     <term name="available at">tersedia pada</term>
-    <term name="by">by</term>
+    <term name="by">oleh</term>
     <term name="circa">circa</term>
     <term name="circa" form="short">c.</term>
     <term name="cited">dikutip</term>
@@ -36,8 +36,8 @@
     <term name="forthcoming">mendatang</term>
     <term name="from">dari</term>
     <term name="ibid">ibid.</term>
-    <term name="in">in</term>
-    <term name="in press">in press</term>
+    <term name="in">dalam</term>
+    <term name="in press">dalam percetakan</term>
     <term name="internet">internet</term>
     <term name="interview">wawancara</term>
     <term name="letter">surat</term>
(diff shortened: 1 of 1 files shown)


locales-id-ID.xml
```diff
@@ -69,13 +69,13 @@
...
 
     <!-- ORDINALS -->
-    <term name="ordinal"></term>
-    <term name="ordinal-01"></term>
-    <term name="ordinal-02"></term>
-    <term name="ordinal-03"></term>
-    <term name="ordinal-11"></term>
-    <term name="ordinal-12"></term>
-    <term name="ordinal-13"></term>
+    <term name="ordinal">bilangan bertingkat</term>
```
adam3smith: I think you meay have misunderstood this? the "ordinal" term refers to what gets added to a number to turn it into an ordinal. So in English for numbers >3 that's "th" -- 4th, 5th, 6th, 7th etc. Unfortunately, we can't do prefixes, if that's how this works in Indonesian?

adam3smith:
@rmzelle @inukshuk  -- how hard would it be to run an instance of Sheldon with slightly different text on the locales? We get far fewer PRs, so only worth it if super easy, but would still be nice.

inukshuk:
There are two options: we could just set up a second Sheldon instance on Heroku and configure it to post to the locales repository. This should not require any code changes at all, but it would break our continuous deployment approach; it also means we'd have to monitor two services. So the better option, I think, is to allow Sheldon to be configured to post to multiple repositories...

Would you want to use exactly the same messages? Or different messages per repository?

locales-id-ID.xml
rmzelle: It is: https://en.wikipedia.org/wiki/Ordinal_indicator#Representation_as_prefix

If we decide we want to support this, perhaps the best option (no pun intended) would be to add a toggle on `<style-options/>`. The only alternatives I can think of, having separate terms (`<term name="ordinal-suffix-01">`) or a term attribute (`<term name="ordinal-01" position="before">`) seem a lot messier.

rmzelle:
Different messages.

As @adam3smith says, it would be nice to have, although the volume of PRs is much lower here than in the CSL style repo. I think that it would indeed be preferable to have a single instance watch two repos.

locales-id-ID.xml
xbypass: No, it is right. If it doesn't support prefix, it should be left as the original for time being. Indonesian as Malay language use prefix to denote ordinal rather than suffix as English does.

Perhaps, someday this will be supported as Malay, Filipino and Chinese used prefix to denote ordinal numbers.

rmzelle: @xbypass, could you use the pencil/edit icon at https://github.com/xbypass/locales/blob/patch-3/locales-id-ID.xml and change the ordinal term translations back from

```xml
    <term name="ordinal">bilangan bertingkat</term>
    <term name="ordinal-01">ke-1</term>
    <term name="ordinal-02">ke-2</term>
    <term name="ordinal-03">ke-3</term>
    <term name="ordinal-11">ke-11</term>
    <term name="ordinal-12">ke-12</term>
    <term name="ordinal-13">ke-13</term>
```

to

```xml
    <term name="ordinal"></term>
    <term name="ordinal-01"></term>
    <term name="ordinal-02"></term>
    <term name="ordinal-03"></term>
    <term name="ordinal-11"></term>
    <term name="ordinal-12"></term>
    <term name="ordinal-13"></term>
```

?

rmzelle: > If it doesn't support prefix, it should be left as the original for time being.

So yes, CSL currently doesn't support this, although it probably should.

adam3smith:
superseded by #166 (which was merged). Closing -- and sorry for letting this linger so long.

PR #131, opened 2016-02-05
Update locales-nl-NL.xml

tomasznguyen:
Changed translation of the term "from" from "van" to "op". "from" is used in the phrase "retrieved from", e.g. in the case of web sites. In Dutch, "retrieved from" is preferred to be translated as "geraadpleegd op".

PR #132, opened 2016-02-05
Update locales-nl-NL.xml

tomasznguyen:
Changed "geraadpleegd" to "geraadpleegd op".

The phrase "Retrieved Year, Month Day, from URL" is in Dutch APA-guidelines often translated as "Geraadpleegd op dag maand jaar, op URL", e.g. in http://itswww.uvt.nl/lis/es/apa/apa-handleiding.pdf. Therefore, I'd like to propose to change "geraadpleegd" to "geraadpleegd op".

### DIFF
--- a/locales-nl-NL.xml
+++ b/locales-nl-NL.xml
@@ -20,7 +20,7 @@
     <date-part name="year"/>
   </date>
   <terms>
-    <term name="accessed">geraadpleegd</term>
+    <term name="accessed">geraadpleegd op</term>
     <term name="and">en</term>
     <term name="and others">en anderen</term>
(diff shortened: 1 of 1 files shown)


adam3smith:
it's going to be easier if you put these in a single pull request rather than multiple. Otherwise, if we accept one of them, the next one can't be merged easily anymore. So why don't you include this change to your other pull request. The instructions for updating pull requests are above.

tomasznguyen:
I apologise. I didn't know that it was possible to edit a previous pull request. I'll update the previous one, so this one can be discarded.

rmzelle:
Continued in #131.

PR #136, opened 2016-03-17
Patch 1

abdealikhurrum:
I noticed that it doesn't have a region suffix...maybe that might be why it's not working?
also, added "mudawwir" for director


rmzelle:
I'm pretty sure that this should remain `locales-ar.xml`

I explicitly changed the file from "ar-AR" to "ar" [https://github.com/citation-style-language/locales/commit/c745e11f0459e76fbe65ff8fe2c1f6ba9eb67cf3] when I realized that "AR" is the region code for Argentina (see http://r12a.github.io/apps/subtags/), and that there is no obvious region for Arabic.

abdealikhurrum:
Oh, sorry about that...

rmzelle:
If you reopen this pull request, and change back `xml:lang="ar-AR"` to `xml:lang="ar"`, we can still accept the translations for the "director" and "by" terms, though.

rmzelle:
Great, thanks!

PR #137, opened 2016-03-19
Update locales-ar.xml

abdealikhurrum:
Since Farsi works and Arabic doesn't, I decided to add what I saw as different between the two (the ordinals - maybe they're required?).

### DIFF
--- a/locales-ar.xml
+++ b/locales-ar.xml
@@ -22,7 +22,7 @@
     <term name="anonymous">مجهول</term>
     <term name="anonymous" form="short">مجهول</term>
     <term name="at">عند</term>
-    <term name="available at">available at</term>
+    <term name="available at">موجود في</term>
     <term name="by">عن طريق</term>
     <term name="circa">حوالي</term>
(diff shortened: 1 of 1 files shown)


locales-ar.xml
```diff
@@ -70,6 +70,12 @@
 
     <!-- ORDINALS -->
     <term name="ordinal"></term>
+    <term name="ordinal-01"></term>
+    <term name="ordinal-02"></term>
+    <term name="ordinal-03"></term>
+    <term name="ordinal-11"></term>
+    <term name="ordinal-12"></term>
+    <term name="ordinal-13"></term>
```
rmzelle: @abdealikhurrum, we'd be happy to accept the translation for the "available at" term, but these empty `<term name="ordinal-xx"></term>` terms are unnecessary and shouldn't make a difference.

rmzelle:
> (the ordinals - maybe they're required?)

No, they're not. If you remove them again I'll accept the other change.

rmzelle:
Thanks!

PR #139, opened 2016-03-27
Update locales-hr-HR.xml

tvrbanec:
U starom prijevodu sve vrvi od srpskih riječi.
The old "translation" is actually serbian.

### DIFF
--- a/locales-hr-HR.xml
+++ b/locales-hr-HR.xml
@@ -22,7 +22,7 @@
     <term name="anonymous">anonim</term>
     <term name="anonymous" form="short">anon.</term>
     <term name="at">na</term>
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks!

tvrbanec:
Thank You? I have two questions:
1. When will I see the actual merging result in my documents?
2. Can I made some more changes in Croatian translation?

rmzelle:
1) That depends on the software you're using this locale file with. E.g. Zotero updates its CSL locale file with every release, but not all programs/products do.
2) Yes, just click the edit button at https://github.com/citation-style-language/locales/blob/master/locales-hr-HR.xml to start editing the most recent version.

PR #140, opened 2016-03-28
Update locales-hr-HR.xml

tvrbanec:
Updated and refined translation.

### DIFF
--- a/locales-hr-HR.xml
+++ b/locales-hr-HR.xml
@@ -70,12 +70,6 @@
 
     <!-- ORDINALS -->
     <term name="ordinal"></term>
-    <term name="ordinal-01"></term>
-    <term name="ordinal-02"></term>
-    <term name="ordinal-03"></term>
-    <term name="ordinal-11"></term>
-    <term name="ordinal-12"></term>
-    <term name="ordinal-13"></term>
 
     <!-- LONG ORDINALS -->
     <term name="long-ordinal-01">prvi</term>
--- a/locales-hr-HR.xml
+++ b/locales-hr-HR.xml
@@ -69,7 +69,7 @@
     <term name="page-range-delimiter">–</term>
 
     <!-- ORDINALS -->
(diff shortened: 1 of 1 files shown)


locales-hr-HR.xml
```diff
@@ -69,13 +69,13 @@
...
-    <term name="ordinal-11">th</term>
-    <term name="ordinal-12">th</term>
-    <term name="ordinal-13">th</term>
+    <term name="ordinal"></term>
+    <term name="ordinal-01"></term>
+    <term name="ordinal-02"></term>
+    <term name="ordinal-03"></term>
+    <term name="ordinal-11"></term>
+    <term name="ordinal-12"></term>
+    <term name="ordinal-13"></term>
```
rmzelle: If there are no ordinal suffixes in Croatian, you can just replace

```
    <!-- ORDINALS -->
    <term name="ordinal"></term>
    <term name="ordinal-01"></term>
    <term name="ordinal-02"></term>
    <term name="ordinal-03"></term>
    <term name="ordinal-11"></term>
    <term name="ordinal-12"></term>
    <term name="ordinal-13"></term>
```

with

```
    <!-- ORDINALS -->
    <term name="ordinal"></term>
```

rmzelle:
Can you make the change I requested (use edit button at https://github.com/tvrbanec/locales/blob/patch-2/locales-hr-HR.xml)? Otherwise it all looks good!

tvrbanec:
A s far as I know :-), there are no ordinals in Croatian...

rmzelle:
> As far as I know :-), there are no ordinal suffixes in Croatian...

Wikipedia lists Croatian as one of the languages that uses an ordinal period (like German): https://en.wikipedia.org/wiki/Ordinal_indicator#Representation_as_period

Does that sound familiar?

rmzelle:
Thanks!

PR #142, opened 2016-04-10
Update locales-zh-CN.xml

rongls:


adam3smith:
Thank you! Given that none of use speaks any Chinese, it'd be helpful if you could briefly explain the nature and reason for this change so we have a record of that down the road.

rongls:
In Chinese, "和" is a more prevalent translation of "and", while "及" is used for the third or more object only. Moreover, "和" is the official translation in academic papers.

rmzelle:
Okay, thanks!

PR #144, opened 2016-05-28
Update locales-it-IT.xml

Pixm79:
Changed two abreviated form ("a cura di") and put quotation marks to standard italian (" ")

### DIFF
--- a/locales-it-IT.xml
+++ b/locales-it-IT.xml
@@ -62,10 +62,10 @@
     <term name="bc">a.C.</term>
 
     <!-- PUNCTUATION -->
-    <term name="open-quote">«</term>
-    <term name="close-quote">»</term>
-    <term name="open-inner-quote">“</term>
-    <term name="close-inner-quote">”</term>
+    <term name="open-quote">“</term>
+    <term name="close-quote">”</term>
+    <term name="open-inner-quote">‘</term>
+    <term name="close-inner-quote">’</term>
     <term name="page-range-delimiter">–</term>
 
     <!-- ORDINALS -->
@@ -229,8 +229,8 @@
       <multiple>dirs.</multiple>
     </term>
     <term name="editor" form="short">
-      <single>a c. di</single>
-      <multiple>a c. di</multiple>
+      <single>a cura di</single>
+      <multiple>a cura di</multiple>
     </term>
     <term name="editorial-director" form="short">
       <single>ed.</single>
@@ -245,8 +245,8 @@
       <multiple>trad.</multiple>
     </term>
     <term name="editortranslator" form="short">
-      <single>a c. di e trad. da</single>
-      <multiple>a c. di e trad. da</multiple>
+      <single>a cura di e trad. da</single>
+      <multiple>a cura di e trad. da</multiple>
     </term>
 
     <!-- VERB ROLE FORMS -->


adam3smith:
What are you basing these on? 
For the quotation marks, we're following the list on Wikipedia, which in turn links to http://www.treccani.it/enciclopedia/discorso-diretto_%28Enciclopedia-dell'Italiano%29/ showing "French" as opposed to "English" style quotation marks. Neither has to be right, but we'd want a solid basis for changing it.

For "a cura di" vs "a c. di" -- @rmzelle you introduced the "c." in https://github.com/citation-style-language/locales/commit/e5b9d1496def6cf1465c3dd6bef0018e589bd02b any recollection where it came from. I am seeing a fair amount of "a c. di" in google searches:
https://www.google.de/search?client=ubuntu&channel=fs&q=%22a+c.+di%22&ie=utf-8&oe=utf-8&gfe_rd=cr&ei=nnFJV8W9MvKA8QfG36fICA&gws_rd=cr#channel=fs&q=%22a+c.+di%22+italiano
and given that you can get a cura di via the long form, I'd tend towards keeping it, but open to be convinced.

adam3smith:
Related: https://forums.zotero.org/discussion/59497/change-quotation-style/#Comment_254719

Pixm79:
Thank you Adam.
It is true that "a c. di" (104.000) can be found in google books, but you will see that is far more common the extensive form "a cura di" (293.000). Moreover, this the the form most commonly require by publishers in Italy.
As for quotation marks, the use of the French version is very old-fashioned in Italy and only few editors still require those. 

For both my changes, you can refer to the style guides of some of the most renowned Publishers in Italy, such as Feltrinelli, Carocci and Bompiani.

adam3smith:
I'm mostly convinced on the quotation marks. Would be good to hear from another Italian academic on this to confirm. @fiapps --what do you think?

But for the a cura di vs. a c. di -- that's different since we don't actually have to choose. As long as both the long form and the short form are in use, we want to have both available. That's why CSL has form="short" and form="long", so I'm pretty sure we don't want to change this.

Pixm79:
Sounds good. As for "a cura di", however, I wasn't able to select th long form: I think (but I might be wrong) that only the short one was listed.

fiapps:
I don't know whether I qualify as an Italian academic, since Italian is not my native language. I do currently live in Italy, and I've written journal articles and conference papers in Italian, but have never supplied with a style guide. I follow rules of style that Italian professors taught me.

I still see «» widely used in Italian. Since I had never looked at a publisher's style guide, I did a Google search for "norme editoriali." The first four results all call for the use of «» for quotations:

  * [Mimesis](http://mimesisedizioni.it/Norme-editoriali-Mimesis.pdf)
  * [Olschki](http://www.olschki.it/media/35dc7390.pdf)
  * [Edizioni ETS](http://www.edizioniets.com/norme_ed.asp): specified for "continental style". Authors are allowed to freely choose between this and "Anglo-Saxon style," in which I suppose “” could be used, although it is not specified. Anglo-Saxon style here is characterized by the use of an author-date citation style.
  * [Arachne](http://www.aracneeditrice.it/scaricabili/norme_redazionali.pdf): "for quotations of more than 5 words." It is not specified what one does for shorter quotations.

The fifth result was a 1992 style guide, which I didn't look at because it could be considered old-fashioned. I didn't look at further search results.

More than one of these specifies a different use for "", such as what are called scare quotes in English.

I conclude that this is a question of style. To me, «» still seems to be the prevalent style, based on my experience and the quick search I conducted, but with the proviso that I am not a native speaker. As a question of style, variations from what is considered prevalent should be made in the particular CSL style that corresponds to a publisher's requirements. I have little personal stake in which form of quotations marks are considered the standard for Italian, since I am using BibLaTeX more than CSL at present, and am quite capable of modifying a style to meet my requirements. Choose what you will as the standard.

adam3smith:
A cura di is available here: https://github.com/citation-style-language/locales/blob/master/locales-it-IT.xml#L255 -- the relevant style defines which one of the two is used. 

I looked at some more style guides and I see the «» almost universally used in academic style (e.g. http://www.insegnadelgiglio.it/wp-content/uploads/2013/06/qda-norme-2015.pdf
http://www.insegnadelgiglio.it/wp-content/uploads/2013/06/universita-venezia-norme.pdf
http://www.paginasc.it/pdf/Norme_EdizionidiPagina.pdf
https://www.francoangeli.it/riviste/NR/Sf-norme.pdf
in addition to those cited above, so I'd be inclined to leave this as is -- with as @fiapps says correctly, the ability to adjust for individual styles.

rmzelle:
> For "a cura di" vs "a c. di" -- @rmzelle you introduced the "c." in e5b9d14 any recollection where it came from.

It's from an earlier pull request from @fiapps which I committed by hand: https://github.com/citation-style-language/locales/pull/31#issuecomment-11946793

rmzelle:
Also, reading over all this, it looks it would be best to just close this pull request. Agreed, @adam3smith? 

(@Pixm79, as @adam3smith mentioned, your proposed changes can always be implemented in the styles themselves, which override the locale files)

adam3smith:
Yes, closing this for the reasons above. Thank you, though, for contributing, and do let us know if there are good reasons for including one or the other of this in a given style.

PR #148, opened 2016-12-14
Update locales-de-DE.xml

tillheilmann:
Added (non-breaking) space to terms "ad" and "bc"

### DIFF
--- a/locales-de-DE.xml
+++ b/locales-de-DE.xml
@@ -58,8 +58,8 @@
     <term name="version">Version</term>
 
     <!-- ANNO DOMINI; BEFORE CHRIST -->
-    <term name="ad">n.&#160;Chr.</term>
-    <term name="bc">v.&#160;Chr.</term>
+    <term name="ad">&#160;n.&#160;Chr.</term>
+    <term name="bc">&#160;v.&#160;Chr.</term>
 
     <!-- PUNCTUATION -->
     <term name="open-quote">„</term>


adam3smith:
@rmzelle -- just want to confirm with you that you agree this is right in the locale (I told Till it is): The English 40bc is 40 v. Chr. in German, so it makes sense to me to code the space into the term.

rmzelle:
@adam3smith, do most style guides not use a separating space? E.g. https://en.wikipedia.org/wiki/Anno_Domini does ("AD 525" and "1 BC")

@tillheilmann, have you confirmed that this change actually works? (in Zotero, Mendeley, or something else)

adam3smith:
you're right. Chicago Manual adds a space, too. @fbennett do you have any recollection on why citeproc-js doesn't add a space between year and BC/AD terms?

rmzelle:
> do you have any recollection on why citeproc-js doesn't add a space between year and BC/AD terms?

Maybe because we wrote it that way in the spec? :)

http://docs.citationstyles.org/en/stable/specification.html#ad-and-bc says

> The “ad” term (Anno Domini) is automatically appended to positive years of less than four digits (e.g. “79” becomes “79AD”). The “bc” term (Before Christ) is automatically appended to negative years (e.g. “-2500” becomes “2500BC”).

As long as the CSL processors don't trim white-space characters defined in the term definitions, this is probably the right solution, since it allows for the choice between regular and non-breaking delimiting spaces, as proposed in this PR.

adam3smith:
good point -- we're taking these then, yes?

rmzelle:
Could wait for Frank to chime in, but otherwise yes.

tillheilmann:
@rmzelle, the change of terms (with added non-breaking spaces) works for me if I redefine the terms in the header of my custom CSL files and produce output with Zotero (4.0) on my Mac. I have not tried other versions/programs/OSs.

rmzelle:
Thanks!

PR #152, opened 2017-01-16
Add badges to README.md

zuphilip:


rmzelle:
Also see https://github.com/citation-style-language/journals/pull/24 and https://github.com/citation-style-language/styles/pull/2432

rmzelle:
Thanks! Now we just have to remember to update the counter when we add additional locales.

zuphilip:
Yes! On the other side, nobody will check the count accurately 🛂 and the number is already high to convince 🎯 ;-)

rmzelle:
Convince darts/targets?

PR #153, opened 2017-01-18
Update locales-lt-LT.xml

embar-:


rmzelle:
Hi @embar-, you probably still want to assign a gender to certain terms? Right now none of the masculine or feminine ordinals will ever be used. You can assign gender to the month term and any of the number variables (see http://docs.citationstyles.org/en/stable/specification.html#gender-specific-ordinals).

embar-:
It is rare cases then discrepancies could be noticed. We can accept as it is.

rmzelle:
> It is rare cases then discrepancies could be noticed. We can accept as it is.

But it doesn't make much sense to define masculine and feminine ordinal terms in the locale file, but not assign any genders to terms that actually invoke the use of those masculine/feminine ordinals.

Our tests failed because you assigned genders to terms associated with variables that can't currently be rendered in ordinal form through `<number/>`. I would greatly prefer it if you would assign genders to the terms that are allowed (those with matching names of the [number variables](http://docs.citationstyles.org/en/stable/specification.html#number-variables), and the month terms; see https://github.com/embar-/locales/blob/master/locales-fr-FR.xml for an example). Or, alternatively, remove the masculine and feminine ordinal terms.

rmzelle:
@aurimasv, you speak Lithuanian, right? Can you provide any input on whether it would be good to add gender-specific ordinal terms to the Lithuanian locale file?

aurimasv:
Gender-specific ordinals certainly make sense. I can comment on which terms are which gender, or submit a follow-up PR if you want.

rmzelle:
> I can comment on which terms are which gender, or submit a follow-up PR if you want.

You could also create a PR against this one. (i.e. https://github.com/embar-/locales/blob/patch-1/locales-lt-LT.xml)

embar-:
Now I looked at Latvian 
(<term name="edition" gender="feminine">) 
and Russian  
(<term name="issue" gender="masculine">
<term name="number-of-pages" gender="feminine">
<term name="volume" gender="masculine">)
examples. 
But in Lithuanian relevant terms all are in masculine. Thus I remove feminine.

<term name="ordinal-01">-iasis</term> was mistake: 
-iasis correct is only for 3-iasis, 23-iasis, 33-iasis, ... 93-iasis, 103-iasis, 123-iasis, ... 193-iasis etc, 
except *13-asis (13-asis, 113-asis, 213-asis etc).
I am not sure how to correctly define this ruler

locales-lt-LT.xml
```diff
@@ -77,6 +75,7 @@
 
     <!-- ORDINALS -->
     <term name="ordinal">-asis</term>
+    <term name="ordinal-03">-iasis</term>
```
aurimasv: You can specify "ordinal-13" to override this further for the 13 case. I think it was worthwhile to keep the feminine ordinals. It may not be used currently, but it doesn't hurt to have it in here.

aurimasv:
@rmzelle, which nouns need genders? I can annotate all of them, but it seems like most of them don't need gender annotations.

locales-lt-LT.xml
```diff
@@ -169,7 +168,7 @@
     <term name="issue" form="short">nr.</term>
     <term name="line" form="short">l.</term>
```
aurimasv: eil.

locales-lt-LT.xml
```diff
@@ -29,22 +27,22 @@
...
       <single>leidimas</single>
       <multiple>leidimai</multiple>
     </term>
     <term name="edition" form="short">leid.</term>
-    <term name="et-al">et al.</term>
+    <term name="et-al">ir kt.</term>
     <term name="forthcoming">ruošiamas</term>
     <term name="from"></term>
-    <term name="ibid">ibid.</term>
+    <term name="ibid">ten pat</term>
```
aurimasv: I've seen ibid used in Lithuanian citations, but I don't know what's more common, just pointing this out. e.g. http://www.llti.lt/failai/SLL22_Str_Jakonyte(1).pdf

embar-:
created clean pull request
https://github.com/citation-style-language/locales/pull/154

locales-lt-LT.xml
```diff
@@ -258,13 +257,13 @@
 
     <!-- VERB ROLE FORMS -->
     <term name="container-author" form="verb"></term>
-    <term name="director" form="verb">directed by</term>
+    <term name="director" form="verb">vadovavo</term>
```
aurimasv: Also change the short form below

locales-lt-LT.xml
```diff
@@ -258,13 +257,13 @@
...
     <term name="editor" form="verb">sudarė</term>
     <term name="editorial-director" form="verb">parengė</term>
-    <term name="illustrator" form="verb">illustrated by</term>
+    <term name="illustrator" form="verb">iliustravo</term>
```
aurimasv: Also fix short form below.

locales-lt-LT.xml
```diff
@@ -169,7 +168,7 @@
     <term name="issue" form="short">nr.</term>
     <term name="line" form="short">l.</term>
     <term name="note" form="short">n.</term>
```
aurimasv: pstb.? I actually have no idea what the proper abbreviation should be, but this doesn't match the long form.

rmzelle:
> which nouns need genders?

Per https://github.com/citation-style-language/locales/pull/153#issuecomment-273944521, just the terms matching the names of the CSL number variables, and the (long form) month terms.

embar-:
In Lithuanian usually we don't write letters at days (like `sausio 1-oji diena`) in official standards. In oficial standards we use: `2016-01-01` or `2016 m. sausio 1 d.`

If in some style could happen, then feminine terms would be for:
1-oji dalis (eng. "part")
1-oji knyga (eng. "book")
1-oji skiltis (eng. "column")
1-oji pastaba (eng. "note")
1-oji eilutė (eng. "line")

locales-lt-LT.xml
embar-: Both `ibid` (Latin form) and `ten pat` (Lithuanian) are allowed and can be found.

locales-lt-LT.xml
embar-: Fixed for 154

locales-lt-LT.xml
embar-: Fixed for 154

locales-lt-LT.xml
embar-: Short form already was corrected.

locales-lt-LT.xml
embar-: Short form also already was correct.

locales-lt-LT.xml
embar-: I never saw short form for this

aurimasv:
> note that days in ordinal form get their gender from the month

Interesting. In Lithuanian, the gender of the day ordinal would derive from "day", which is feminine, irrespective of the month. But as @embar- says, days don't usually use ordinals in Lithiuanian citations, but if we do want to support that case, all months would need to be labeled feminine.

locales-lt-LT.xml
aurimasv: I mean the verb form below (line 271), which (on this pull request) still says "dir.". Illustrator says "illus."

embar-: You are right, previous I looked at 234 and 246 lines with same short forms.
Fixed in https://github.com/citation-style-language/locales/pull/154/commits/dfea4a9b0cb1eb136f251cd0f02cc07a628b7413

PR #155, opened 2017-01-22
Update locales-lt-LT.xml with genders

rmzelle:
Per
https://github.com/aurimasv/locales/commit/af03dac4ec35a7b54fffc4ebd9997
ae86ead8c1c and https://github.com/citation-style-language/locales/pull/153

### DIFF
--- a/locales-lt-LT.xml
+++ b/locales-lt-LT.xml
@@ -36,31 +36,31 @@
       <single>leidimas</single>
       <multiple>leidimai</multiple>
     </term>
-    <term name="edition" form="short" gender="masculine">leid.</term>
+    <term name="edition" form="short">leid.</term>
     <term name="et-al">ir kt.</term>
     <term name="forthcoming">ruošiamas</term>
     <term name="from"></term>
     <term name="ibid">ten pat</term>
     <term name="in"></term>
     <term name="in press">priimta spaudai</term>
-    <term name="internet" gender="feminine">prieiga per internetą</term>
-    <term name="interview" gender="masculine">interviu</term>
-    <term name="letter" gender="masculine">laiškas</term>
+    <term name="internet">prieiga per internetą</term>
+    <term name="interview">interviu</term>
+    <term name="letter">laiškas</term>
     <term name="no date">sine anno</term>
     <term name="no date" form="short">s.a.</term>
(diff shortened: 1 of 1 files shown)


rmzelle:
@embar-, @aurimasv, can you review this? I took @aurimasv's version, and then removed anything that wasn't valid CSL 1.0.1 (per http://docs.citationstyles.org/en/stable/specification.html#gender-specific-ordinals, it's only allowed to specify a gender for the long form of CSL number variable terms and months terms).

Apart from the gender assignments, the ordinal terms are changed, and a translator for "illustrator".

embar-:
Gender definitions are good. But as said in https://github.com/citation-style-language/locales/pull/153#issuecomment-273983851 and in https://github.com/citation-style-language/locales/pull/153#issuecomment-273985371  , we don't need genders for dates, but it will be used, then gender of day (as digit) depends on word "diena" (eng. "day", feminine), not on month (mixed genders).

We should keep "ordinal-03" and "ordinal-13" definitions in 78-79 lines.
At 275 line we should keep "iliustr."; see foe example http://internet.unib.ktu.lt/chemija_lt/suda_lite.htm

embar-:
We could also add 
`<term name="ordinal-03" gender-form="masculine">-iasis</term>`
`<term name="ordinal-13" gender-form="masculine">-asis</term>`
`<term name="ordinal-03" gender-form="feminine">-ioji</term>`
`<term name="ordinal-13" gender-form="feminine">-oji</term>`

rmzelle:
How about now?

embar-:
Thanks, this is very good!

rmzelle:
Great. I'll merge this once I get the okay from @aurimasv as well. (Sorry if I was a bit slow in catching up; I don't know anything about Lithuanian)

I removed the genders from the month terms, since it would just lead to confusion, because CSL currently assumes that the day number in ordinal form follows the gender of the month, which you indicated is not the case for Lithuanian. We could assign the "feminine" gender to all "month" terms to hack around this (since "diena" is feminine), but that's also confusing, so I rather leave things as they are. If anybody really would like proper support, we should amend the CSL specification instead.

aurimasv:
Looks good. Thanks, Rintze, @embar-!

PR #157, opened 2017-02-06
Update locales-fa-IR.xml

hamedheydari:
Hi. I'm Sorry for second edition in a short time period.
Also, I would like to ask you three questions:
1- Is it possible to localize the letters of disambiguation of citations? e.g. ([Authors], [Year][**a**]) >>> ([Authors], [Year]_[**الف**])
2- Is it possible to localize punctuation marks? e.g. [**,**] [**;**] >>> [**،**] [**؛**]
3- Mendeley does not support loading CSL locale files. Is it right?

### DIFF
--- a/locales-fa-IR.xml
+++ b/locales-fa-IR.xml
@@ -6,13 +6,13 @@
   </info>
   <style-options punctuation-in-quote="true"/>
   <date form="text">
-    <date-part name="month" suffix=" "/>
-    <date-part name="day" form="numeric-leading-zeros" suffix=", "/>
+    <date-part name="day" suffix=" "/>
+    <date-part name="month" form="numeric-leading-zeros" suffix=" "/>
     <date-part name="year"/>
   </date>
   <date form="numeric">
-    <date-part name="month" form="numeric-leading-zeros" suffix="/"/>
     <date-part name="day" form="numeric-leading-zeros" suffix="/"/>
+    <date-part name="month" form="numeric-leading-zeros" suffix="/"/>
     <date-part name="year"/>
   </date>
   <terms>
@@ -58,8 +58,8 @@
     <term name="version">نسخه</term>
 
     <!-- ANNO DOMINI; BEFORE CHRIST -->
-    <term name="ad">AD</term>
-    <term name="bc">BC</term>
+    <term name="ad">بعد از میلاد</term>
+    <term name="bc">قبل از میلاد</term>
 
     <!-- PUNCTUATION -->
     <term name="open-quote">“</term>
@@ -141,8 +141,8 @@
       <multiple>قسمت‌های</multiple>
     </term>
     <term name="sub verbo">
-      <single>sub verbo</single>
-      <multiple>sub verbis</multiple>
+      <single>در ذیلِ واژه</single>
+      <multiple>در ذیلِ واژه‌های</multiple>
     </term>
     <term name="verse">
       <single>بیت</single>
@@ -160,8 +160,8 @@
     <term name="figure" form="short">تصویر</term>
     <term name="folio" form="short">برگ</term>
     <term name="issue" form="short">ش</term>
-    <term name="line" form="short">l.</term>
-    <term name="note" form="short">n.</term>
+    <term name="line" form="short">خط</term>
+    <term name="note" form="short">یادداشت</term>
     <term name="opus" form="short">قطعه</term>
     <term name="page" form="short">
       <single>ص</single>
@@ -203,12 +203,12 @@
       <multiple>directors</multiple>
     </term>
     <term name="editor">
-      <single>ویرایشگر</single>
-      <multiple>ویرایشگران</multiple>
+      <single>ویراستار</single>
+      <multiple>ویراستاران</multiple>
     </term>
     <term name="editorial-director">
-      <single>ویرایشگر</single>
-      <multiple>ویرایشگران</multiple>
+      <single>ویراستار</single>
+      <multiple>ویراستاران</multiple>
     </term>
     <term name="illustrator">
       <single>طراح گرافیک</single>
@@ -219,8 +219,8 @@
       <multiple>مترجمین</multiple>
     </term>
     <term name="editortranslator">
-      <single>ویرایشگر و مترجم</single>
-      <multiple>ویرایشگران و مترجمین</multiple>
+      <single>ویراستار و مترجم</single>
+      <multiple>ویراستاران و مترجمین</multiple>
     </term>
 
     <!-- SHORT ROLE FORMS -->
@@ -229,12 +229,12 @@
       <multiple>dirs.</multiple>
     </term>
     <term name="editor" form="short">
-      <single>ویرایشگر</single>
-      <multiple>ویرایشگران</multiple>
+      <single>ویراستار</single>
+      <multiple>ویراستاران</multiple>
     </term>
     <term name="editorial-director" form="short">
-      <single>ویرایشگر</single>
-      <multiple>ویرایشگران</multiple>
+      <single>ویراستار</single>
+      <multiple>ویراستاران</multiple>
     </term>
     <term name="illustrator" form="short">
       <single>تصویرگر</single>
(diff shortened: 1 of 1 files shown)


rmzelle:
> Hi. I'm Sorry for second edition in a short time period.

No problem!

> Also, I would like to ask you three questions:
> 1- Is it possible to localize the letters of disambiguation of citations? e.g. ([Authors], [Year][a]) >>> ([Authors], [Year]_[الف])

No. The current scheme ("a" through "z", then "aa" through "zz", etc.) is hard-coded. What kind of scheme do you wish to use?

> 2- Is it possible to localize punctuation marks? e.g. [,] [;] >>> [،] [؛]

There are currently only a few punctuation marks that can be localized. The delimiters used for dates (like the one you changed in this pull request), and quotation marks and the symbol used for page ranges (see https://github.com/citation-style-language/locales/blob/74a9e7975986c119c20454a3be2d973c45cf08a1/locales-en-US.xml#L64).

That said, you can create custom styles for Farsi and change most punctuation as required (CSL currently has a few hard-coded commas, I think; please report any issues you run into here).

> 3- Mendeley does not support loading CSL locale files. Is it right?

No, as far as I know Mendeley periodically updates the CSL locale files it uses in new releases of Mendeley Desktop (not necessarily with every release), and I don't think it's possible yet for users to update the CSL locales by themselves. (see https://github.com/citation-style-language/styles/issues/880#issuecomment-37294212; the Mendeley developer who replied to me in that thread, cpina, is on sabbatical, though, so Mendeley might be updating its CSL locale files even less frequently than normal).

hamedheydari:
<date form="text">
    <date-part name="day" suffix=" "/>
    <date-part name="month" **form="numeric-leading-zeros"** suffix=" "/>

I Forgot to delete the bold section.
-------------------------------------------------------------------

Thank you for your prompt reply.

> No. The current scheme ("a" through "z", then "aa" through "zz", etc.) is hard-coded. What kind of scheme do you wish to use?
الف ب پ ت  ث ... instead of a b c d e ...
https://en.wikipedia.org/wiki/Persian_alphabet
And do you think there is any way to add an underscore between year and year-suffix (e.g. alphabet letters)?
> You can create custom styles for Farsi and change most punctuation as required
OK. So I can create a separate style for Farsi. Although I can't localize the delimiter used for publisher-place. e.g. Tehran**,** Iran > تهران**،** ایران

rmzelle:
> I forgot to delete this section: form="numeric-leading-zeros"

You don't want leading zeros for both the "text" and "numeric" formats?

> '... الف ب پ ت ث' instead of 'a b c d e etc'
> (https://en.wikipedia.org/wiki/Persian_alphabet)

Do you know what happens when you run out of letters? (as I mentioned above, in English, you normally would go from "x", "y", "z", to "aa", "ab", ac", etc., to "ba", "bb", "bc", etc.)

rmzelle:
> And do you think there is any way to add an en dash (Alt+0150: – ) between year and year-suffix (alphabet letters)?

Yes. By default, the year-suffix is placed automatically, but you can change its position (and surrounding punctuation) by using something like `<text variable="year-suffix" prefix="–"/>`.

> However I can't localize the delimiter used for publisher-place. e.g. [Tehran, Iran] تهران، ایران <<< تهران, ایران. And Mendeley MS Word Plugin doesn't support two style in a document.

CSL currently sees "publisher-place" as a dumb string (it doesn't recognize that the field contains "Tehran" and "Iran" separated by ", ", but instead just sees "Tehran, Iran"), so whatever is in the field is what's printed. Do you need both the English and Farsi translations within the same reference?

hamedheydari:
> You don't want leading zeros for both the "text" and "numeric" formats?
>
I don't want that only for text format.
> Do you know what happens when you run out of letters?
> 
Yes. We can use a same algorithm in Persian. 
"ِالف"، "بِ"، "پِ"، "تِ"، "ثِ"، "جیم"، "چِ"، حِ"، "خِ"، "دال"، "ذال" ... "میم"، "نون"، "واو"، "هِ"، "ی" ←
then: "ِالف الف"، "الف بِ"، "الف پ" ... and then: "ِبِ الف"، بِ بِ"، "بِ پ" ...
Although in fact we need just two or three letters and 32 letters of Persian alphabet would be more than enough.
> By default, the year-suffix is placed automatically, but you can change its position. by using something like `<text variable="year-suffix" prefix="–"/>`
> 
Thanks. Done. How about text direction (for Bibliography)?
> Do you need both the English and Farsi translations within the same reference?
> 
No. I mean, sometimes we have English references; sometimes we have Farsi references (entirely Farsi); sometimes we have both of them and should be list in a document. and we can't use two style in a document. That's a problem related to Mendeley.

rmzelle:
>> You don't want leading zeros for both the "text" and "numeric" formats?
> 
> I don't want that for text format.

Fixed: https://github.com/citation-style-language/locales/commit/b01a5e4d420a4c0e446625c271144c52b0ab0a76

> Yes. We can use a same algorithm in Persian.

Okay, great. Issue created, although it might be a while before we can add support for non-English year suffixes.

> How about RTL Text Direction? (for Bibliography in MS Word, as default). It does change to LTR, if I add a citation.

What happens, exactly? The whole citation changes to LTR whenever there is year-suffix disambiguation?

> No. I mean, sometimes we have English references; sometimes we have Farsi references (entirely Farsi); sometimes we have both of them and should be list in a document. and we can't use two style in a document. I think that's a problem related to Mendeley.

That's a limitation in CSL that we're aware of. It's a priority to add support for references in different languages in a single bibliography, but again it might take a while.

hamedheydari:
Thank you so much.

> What happens, exactly? The whole citation changes to LTR whenever there is year-suffix disambiguation?
> 
I made a short video to explain this problem. (20s/ 804kb) . you can watch it at:
https://www.dropbox.com/s/lnyzpe0be4z41km/Mendeley-RTL.mp4?dl=0

rmzelle:
> I made a short video to explain this problem. (20s/ 804kb).
> https://www.dropbox.com/s/lnyzpe0be4z41km/Mendeley-RTL.mp4?dl=0

So any time a citation is inserted, the bibliography becomes LTR?

hamedheydari:
@ rmzelle Yes, exactly. 
and last my question (!): how csl differentiates 'number of pages' from 'pages'? I guess If 'type: book' > number of pages; if else > pages.

adam3smith:
CSL is agnostic about pages/number-of-pages. You can have both for the same item, in theory, though that'd be highly unusual in practice. In Mendeley, it does indeed look like it behaves like you say. In Zotero, for example, "Pages" and "# of Pages" are differently named fields, which I think is a better solution.

For your RTL-->LTR issue, in the end that's Mendeley, not CSL, too, but see if Mendeley applies a Word style (see https://support.office.com/en-us/article/Customize-styles-in-Word-D38D6E47-F6FC-48EB-A607-1EB120DEC563 ) to the bibliography. You may be able to fix that to RTL.

rmzelle:
> How does CSL (or Mendeley) differentiate 'number of pages' from 'pages'?

CSL has two separate variables, "number-of-pages" and "page", but Mendeley might only be using the latter (incorrectly).

rmzelle:
@fbennett, do I recall correctly that you did some work to get RTL to work correctly in citeproc-js? Could you take a brief look at https://github.com/citation-style-language/locales/pull/157#issuecomment-279197754, and tell us whether you think this is a problem in citeproc-js or Mendeley? Does citeproc-js automatically apply RTL for styles used with RTL locales?

hamedheydari:
@adam3smith @rmzelle Thank You.
Apparently Bibliograpghy Text Direction is dependent on 'Normal' style in Word. Because it becomes reverse, if I change that. (For English references, with any bibliography refresh, it becomes RTL!)

fbennett:
@rmzelle There is some (long-stale) code in the processor that attempts to force RTL on items set to certain languages (`["ar", "he", "fa", "ur", "yi", "ps", "syr"]`), using Unicode characters that (theoretically) impose text direction. It is controlled by a processor toggle that is currently set `false` by default. If I remember correctly, Word and LibreOffice handle RTL inserts differently, and at the time that code was written (years ago), we weren't able to get the output correct in both environments. We could try turning the toggle on in a Propachi variant to see if it helps in @hamedheydari's document ... ?

rmzelle:
> We could try turning the toggle on in a Propachi variant to see if it helps in @hamedheydari's document ... ?

@hamedheydari is using Mendeley, so that wouldn't work.

> Apparently Text Direction for Bibliograpghy is dependent on 'Normal' style in Word. (Just 'Normal' style, not others even though that was selected). Because that issue happens oppositely, if I change Text Direction in 'Normal' style. (English references becomes RTL with any bibliography refresh)

Ah, yes, if you have mixed references in RTL and LTR languages in a single bibliography, this can't be fixed by modifying the word processor style.

fbennett:
I'll write to the developers there with the suggestion, so they can pick it up.

rmzelle:
> I'll write to the developers there with the suggestion, so they can pick it up.

You have good contacts at Mendeley?

fbennett:
Just @cpina (hi, Carles, welcome to this issue :)

adam3smith:
Carles is on sabbatical; specifically (and I'm not making this up) on a boat circumnavigating antarctica.

fbennett:
Oh yes, of course. How could I have forgotten that.

PR #158, opened 2017-03-17
Translation updates sl-SI

KrOstir:
E.g. terms accessed, available at, et-al; corrected some plural versions.


rmzelle:
@ratek1, could you give this a quick look-over?

ratek1:
It looks good to me (corrected translations).

rmzelle:
Great, thanks both.

PR #160, opened 2017-04-03
Update locales-pl-PL.xml

pa0:
improvements


rmzelle:
@MichauSm, if you're still around, could you take a quick look and tell us whether these changes look correct to you?

(@pa0, we usually ask a previous contributor to double-check any changes to locale files)

MichauSm:
@rmzelle, yes, still aroud :) contribution of @pa0 is ok.

rmzelle:
Great, thanks both!

PR #161, opened 2017-05-31
Add composer.json

stucki:
Simplify integration of this repository through Composer.

However, as long as the package is missing on https://packagist.org/, the repository still needs to be added to the `composer.json` file of a project.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,19 @@
+{
+    "name": "citation-style-language/locales",
+    "description": "Citation Style Language (CSL) Locales",
+    "type": "library",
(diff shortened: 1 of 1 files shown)


stucki:
@seboettg This should be useful for https://packagist.org/packages/seboettg/citeproc-php

adam3smith:
@rmzelle these are fine by me. Any concerns from your end? Do you know if non-csl files on csl-distribution get automatically synced over? (I think the answer is no, but wanted to check)

hxdef:
@adam3smith Sorry for late response. Checked your merge, see no any issues with that. It should be compatible with the github action. Thanks!

PR #165, opened 2017-10-03
Update locales-ro-RO.xml

nturcan:
I translated correctly editorial-director as „coordonator”.


adam3smith:
Ditto here -- thanks!

PR #167, opened 2017-11-16
change date suffix comma to arabic comma

abdealikhurrum:
Is the latin comma necessary for dates? I feel the Arabic comma would be more homogenous.


rmzelle:
(continued in #168)

PR #168, opened 2017-11-18
update locales-ar for grammar and completeness

abdealikhurrum:
Changed delimiter in dates from Latin comma (,) to Arabic comma (،)
Changed tense of roles from modified to _marfu'_ (The form they were previously in - ending in ين- is used when that word has been modified by another word or is the object of a verb. The ون form is what it would take if it has been unmodified or at the beginning of a sentence - which is what the use case would be here.)
Added _kashida_ to short forms which have been abbreviated. 
Changed pages (single and multiple) to صـ (I don't think ص.ص. might be appropriate though it is analogous to pp.)
Completed a few entries. (They might not be totally in agreement with what is commonly used, but they can be modified in individual styles according to local tastes or changed here. But I think minimizing the mixing of LTR and RTL characters is very important...it can lead to serious confusion!)

PR #170, opened 2017-12-18
Update locales-ar.xml

gotoayman:
in editing Books, we Arabs use the word "تحقيق"  not "تحرير" 
you almost can find this word "تحقيق" at the top page of every edited book.
thank you.

### DIFF
--- a/locales-ar.xml
+++ b/locales-ar.xml
@@ -199,12 +199,12 @@
 
     <!-- LONG ROLE FORMS -->
     <term name="director">
-      <single>مدور</single>
-      <multiple>مدورين</multiple>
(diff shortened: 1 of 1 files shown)


rmzelle:
@gotoayman, could you check if any of your proposed changes here (https://github.com/citation-style-language/locales/pull/170/files) still need to be applied to https://github.com/citation-style-language/locales/blob/master/locales-ar.xml now that we have accepted the changes from #168? If so, please create a new pull request of off https://github.com/citation-style-language/locales/blob/master/locales-ar.xml.

rmzelle:
Replaced by #172.

PR #171, opened 2017-12-19
Create locales-de–CH–HAFL

kevkev770:
This locale file adapts de locales-de-CH to the BFH-HAFL (Bern University of Applied Sciences School for Agriculture, Forestry and Food Science) requirements. I changed some translations.

### DIFF
--- a/locales-de–CH–HAFL
+++ b/locales-de–CH–HAFL
@@ -1,0 +1,227 @@
+<?xml version="1.0" encoding="utf-8"?>
+<<<<<<< Local Changes
+<style class="in-text" version="1.0" and="text" demote-non-dropping-particle="never" default-locale="de-CH" xmlns="http://purl.org/net/xbiblio/csl">
+=======
+<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0" and="text" demote-non-dropping-particle="never" default-locale="de-CH">
+>>>>>>> External Changes
+  <!-- This style was edited with the Visual CSL Editor (http://editor.citationstyles.org/visualEditor/) -->
+  <info>
+    <title>HAFL_Elsevier_Harvard_2_1</title>
+    <title-short>HAFL2_1</title-short>
+    <id>http://www.zotero.org/styles/hafl_elsevier_harvard_2_1</id>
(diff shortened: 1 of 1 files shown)


rmzelle:
@kevkev770, what you're submitting here is a CSL style, not a CSL locale file. Styles need to be submitted at https://github.com/citation-style-language/styles (see https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md#3a-submitting-a-new-style). Also make sure your style is valid CSL (you need at least replace lines 2-6 in your current submission with either line 3 or 5) (see https://github.com/citation-style-language/styles/wiki/Style-Requirements).

Also, is there an online description of the citation format used by the Bern University of Applied Sciences School for Agriculture, Forestry and Food Science? We generally only accept styles for which some documentation exists (either in print or online).

kevkev770:
Ok thank you for the answer!

From your description I assume I uploaded the wrong code, as I really did change some expressions in the the de-CH locale file. But no worry, I was in need of a faster solution so I just really adapted an existing CSL style and put all the locale code lines in there. That works for now though it’s not really elegant. But However I’m a geographer and not e computer specialist ;-).

Maybe one day I will create the BFH HAFL locale file which is an odd mixture between en-GB and de-CH locale.

Happy Christmas

Best wishes,
Kevin

__________________________________________________________________
kevin suhr | fährweg 47 4600 olten | kevin.suhr@ensemen.com

rmzelle:
> But no worry, I was in need of a faster solution so I just really adapted an existing CSL style and put all the locale code lines in there. That works for now though it’s not really elegant.

That's actually how you would typically do this.

> But However I’m a geographer and not e computer specialist ;-).

(the original creator of CSL, Bruce D'Arcus, also happens to be a geographer by trade. The CSL project is primarily run by researchers without a particular strong background in computer science, so you're in good company :) )

> Maybe one day I will create the BFH HAFL locale file which is an odd mixture between en-GB and de-CH locale.

Just note that we wouldn't accept such a locale file here. CSL is set up in such a way that we maintain a single locale file per language or dialect. We try to use translations in these files that are generally used and accepted, and any style-specific customization must then take place in the style itself.

PR #173, opened 2017-12-29
Add translators from git log to the locale files.

gregoa:
For all xml files without any `<translator>` tag, the authors of the commits
were extracted from git log, and the ones from technical commits / mass
commits over the whole repo manually removed. The rest were added as
`<translator>`s to the xml files.

Background: issue #151
This still leaves 8 files without any `<translator>`s:
```
locales-af-ZA.xml
locales-bg-BG.xml
locales-cy-GB.xml
locales-km-KH.xml
locales-ko-KR.xml
locales-mn-MN.xml
locales-sr-RS.xml
locales-uk-UA.xml
```

PR #174, opened 2018-01-18
Add CONTRIBUTING.md and .github/PULL_REQUEST_TEMPLATE.md

koppor:
This adds initial contribution guidelines. This refs https://github.com/citation-style-language/locales/pull/173 and should ensure that future updates come with a complete list of translators.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,3 @@
+<!-- describe the changes you have made here: what, why, ... -->
+
+- [ ] Checked that author is listed as `<translator>` in the `<info>` block at the beginning of the file.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,13 @@
-<!-- describe the changes you have made here: what, why, ... -->
+## CSL Locales Pull Request Template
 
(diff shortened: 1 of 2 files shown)


koppor:
@adam3smith Does this also look OK for you?

rmzelle:
I'll rewrite this a little bit before merging, and it might be a few days before I have time to do so, but generally this looks good to me.

adam3smith:
Yes, I think we should add to the checklist (formatting and validation at least) but generally I have no objections. @rmzelle feel free to merge when you´re done editing

adam3smith:
ping @rmzelle -- I think this would be nice to have

rmzelle:
> it might be a few days before I have time to do so

*a few days*

Anyway, @adam3smith, let me know if you like it like this. Also, in the many years since we moved to GitHub, I feel that the GitHub wikis have mostly been replaced by README.md and other repository files for the purpose of providing information to users and contributors. It probably makes sense for us to make README.md the starting point for linking to the various sources of documentation (here and in the other main CSL repos).

adam3smith:
Good to merge from my end -- thanks!

rmzelle:
(and do you agree on the wiki sentiment? I think the big advantage of in-repo files is that we get all the benefits of the regular GitHub code workflow, whereas wikis are either locked down or open-to-all)

rmzelle:
Thanks both!

PR #176, opened 2018-02-16
Update locales-bg-BG.xml

valeriina:
all fields translated
some fields corrected
deleted some forms for single and plural case - where it needed

### DIFF
--- a/locales-bg-BG.xml
+++ b/locales-bg-BG.xml
@@ -3,10 +3,6 @@
   <info>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
     <updated>2018-02-15T04:14:02+00:00</updated>
-    <contributor>
-      <name>Valeriya Simeonova</name>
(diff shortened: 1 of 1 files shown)


locales-bg-BG.xml
```diff
@@ -204,22 +228,16 @@
     </term>
 
     <!-- LONG ROLE FORMS -->
-    <term name="director">
-      <single>director</single>
-      <multiple>directors</multiple>
-    </term>
+    <term name="director">под общата редакция на</term>
```
adam3smith: I think this is a misunderstanding of the term -- this director as in movie director. Shouldn't that be режисьор ?

adam3smith:
@valeriina sorry this took forever to review. In general looks great, but I have a question about the translation of "director" -- would you mind taking a look?

valeriina:
Hello everyone,
If the template is for books, the given term is correct (под общата редакция на), if is about art and movies then the appropriate term is режисьор.

All the best!
Valeeiya Simeonova

adam3smith:
Thanks! And apologies for the delay

PR #177, opened 2018-02-28
Update locales-ro-RO.xml

nturcan:
Change page-range-delimiter to n-dash.


adam3smith:
not sure why this never got merged. Sorry about that.

PR #179, opened 2018-04-12
Corrected et-al in sl-SI

KrOstir:


adam3smith:
Thanks! 
(Rintze: KrOstir had submitted the original translation about a year ago, so no need to check with anyone else)

PR #186, opened 2019-01-21
Create Latin translation

adunning:
This is meant to support style files for series such as Corpus Christianorum, <http://www.corpuschristianorum.org/authors.html>. I've translated all the terms, but many will never come up in practice. I'm not aware of a proper Latin citation guide, and would be delighted to see any corrections.

### DIFF
--- a/locales-la-LA.xml
+++ b/locales-la-LA.xml
@@ -1,0 +1,309 @@
+<?xml version="1.0" encoding="utf-8"?>
+<locale xmlns="http://purl.org/net/xbiblio/csl" version="1.0" xml:lang="la-LA">
+  <info>
+    <translator>
+      <name>Andrew Dunning</name>
+    </translator>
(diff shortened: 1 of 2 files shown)


adam3smith:
It's probably not going to see a ton of use, but I'm pretty psyched we have this. Thanks!

adunning:
Thank you! It was quite fun to put together.

rmzelle:
Hey both, I'm a bit late to the party, but it doesn't look like "la-LA" is the correct locale code. Per https://r12a.github.io/app-subtags/, "la-LA" appears to stand for Latin in Laos. Since Latin probably doesn't have a clearly defined geographical area, I suggest trimming the locale code to just "la".

adunning:
I ought to have remembered. Sorry about that!

adunning:
This is fixed in #188.

PR #187, opened 2019-01-21
Italian: Correct ordinal indicator

adunning:
The ° symbol is a degree sign; º is the correct symbol for an ordinal.

### DIFF
--- a/locales-it-IT.xml
+++ b/locales-it-IT.xml
@@ -72,7 +72,7 @@
     <term name="page-range-delimiter">–</term>
 
     <!-- ORDINALS -->
-    <term name="ordinal">°</term>
+    <term name="ordinal">º</term>
 
     <!-- LONG ORDINALS -->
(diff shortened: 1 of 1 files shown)


rmzelle:
@adam3smith, looks correct to me. The diff shows an underlined symbol for me, which seems to be font-specific. See https://en.wikipedia.org/wiki/Ordinal_indicator:

> In Spanish, Portuguese, Italian and Galician, the ordinal indicators º and ª are appended to the numeral depending on whether the grammatical gender is masculine or feminine respectively. These suffixes are often underlined as well, though in regard to digital typography, this will depend on the font used. Cambria and Calibri, for example, have underlined ordinal indicators.
> 
> The masculine ordinal indicator U+00BA (º) is often confused with the degree sign U+00B0 (°)

adam3smith:
Yes, that's unambiguous. Thanks!

PR #188, opened 2019-01-23
Correct Latin locale code

adunning:
Correcting my error in #186.


adunning:
Thank you for catching it! It's doubly embarrassing because I had made a proposal to the IETF a couple of years ago to create Latin variant tags corresponding to the private-use `la-x-classic` and `la-x-liturgic` tags that [TeX Hyphenation](http://www.hyphenation.org/tex#languages) uses. They felt in the event that regional codes were good enough – I've tended to use `la` for classical Latin; `la-VA` for ecclesiastical Latin (if there is such a thing); `la-EU` for neo-Latin; `la-GB` for Anglo-Latin; `la-FR` for Gallo-Latin; etc. But there aren't many occasions on which one needs these.

rmzelle:
Ah, well, in that case I agree with you :). Most of us are not exactly well-versed in the arcane knowledge of locale codes.

PR #192, opened 2019-04-08
Update locales-zh-CN.xml

stultulo:
1. The "于" is used to instead of the "於" in Simplified Chinese.
2. Added a period after the "等".

### DIFF
--- a/locales-zh-CN.xml
+++ b/locales-zh-CN.xml
@@ -38,7 +38,7 @@
     <term name="cited">见引于</term>
     <term name="edition">版本</term>
     <term name="edition" form="short">本</term>
-    <term name="et-al">等.</term>
+    <term name="et-al">等</term>
     <term name="forthcoming">即将出版</term>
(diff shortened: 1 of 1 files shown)


rmzelle:
@sati-bodhi, could you take a look to check whether these changes look correct to you?

redleafnew:
line 28: 见於 should be 见于
line 33: 载於 should be 载于
line 38: 见引於 should be 见引于
line 53: 发表於 should be 发表于。


i.e. all 於 should be replaced with 于。

redleafnew:
Added a period after the "等". is not necessary.

sati-bodhi:
@rmzelle, 
Yes, the changes indicated by @redleafnew and @stultulo are correct.

rmzelle:
> Added a period after the "等". is not necessary.

Thanks all! I removed the period after "等" per @redleafnew's suggestion.

PR #195, opened 2019-05-22
Update locales-nb-NO.xml,locales-nn-NO.xml

kjtlslt:
Short form for "no date" should be u.å. (cf sokogskriv.no, NTNU VIKO, etc.)


rmzelle:
Thanks! Sorry for the long wait.

PR #197, opened 2019-06-11
update sheldon

retorquere:
TBH I'm not *super* happy about the co-dependency of always having to update 3 repos for Sheldon activity, but I really don't see another way.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,8 +1,8 @@
 GIT
   remote: https://github.com/citation-style-language/Sheldon.git
-  revision: ab5d13637fe53da8eab1f924d4708d870cf96391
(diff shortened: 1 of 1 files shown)


rmzelle:
Thanks. Will you document the process of updating Sheldon somewhere (in case a future maintainer takes over from you, or your future self forgot how to do things)?

retorquere:
I've added instructions to https://github.com/citation-style-language/Sheldon/blob/master/README.md - they're perhaps a bit terse, but that's really only what needs to be done.

retorquere:
Come to think of it... there *is* a way to do this without tri-PRs, which would be just to run "bundle update" in `before_script` on Travis.

PR #198, opened 2019-06-12
always fetch latest sheldon

retorquere:
Always fetches the latest sheldon gem in the travis build. This way we don't need to issue PRs in threes when Sheldon updates.


rmzelle:
Thanks!

PR #203, opened 2020-02-27
Fix markdown lint error "[no-literal-urls] Don’t use literal URLs without angle brackets"

koppor:
In JabRef, [we run Codacy](https://devdocs.jabref.org/code-quality) to ensure quote qualty. It uses [remark lint](https://github.com/remarkjs/remark-lint) to lint markdown. Since we include the citation styles in JabRef, it also [complains](https://app.codacy.com/gh/JabRef/jabref/file/42498231068/issues/source?bid=16663292&fileBranchId=16663292#l4) about errors there. One rule is 

    [no-literal-urls] Don’t use literal URLs without angle brackets

grafik

This PR fixes it.

Side comment: Is also complains about two other issues, but I disabled that rules as a) GitHub requires `_` in this filename and b) [markdown-lint](https://github.com/DavidAnson/markdownlint)'s rule is that bulleted lists start at the beginning of a line. I personally like markdownlint more, but it is [not installed at codacy](https://twitter.com/JabRef_org/status/1232752601709633536).

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,7 @@
 ## CSL Locales Pull Request Template
 
 You're about to create a pull request to the CSL locales repository.
-If you haven't done so already, see http://docs.citationstyles.org/en/stable/translating-locale-files.html for instructions on how to translate CSL locale files, and https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md on how to submit your changes.
+If you haven't done so already, see <http://docs.citationstyles.org/en/stable/translating-locale-files.html> for instructions on how to translate CSL locale files, and <https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md> on how to submit your changes.
 In addition, please fill out the pull request template below.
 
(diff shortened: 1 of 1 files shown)


rmzelle:
@koppor, thanks!

Just FYI, for styles (https://github.com/citation-style-language/styles) we set up a downstream repo with all the extra files removed (https://github.com/citation-style-language/styles-distribution). The latter also only gets updated if the Travis CI builds are successful for the former.

We could set up something similar for the "locales" repo, although per https://github.com/citation-style-language/styles/issues/4474 I'm hoping for a simpler setup for both the styles and locales files:

* keep the "master" branches for the most current version of CSL (1.0.1 right now), where we accept external contributions
* create a branch within each repo for the current CSL version ("1.0.1") that fulfills the role of our "styles-distribution" repo and which get automatically updated with the style/locale files from "master" on each successful Travis CI build (we currently only have archival branches, for old CSL versions like 1.0 and 0.8.1).

This would eliminate having secondary repos, and create stable branches for each CSL version that downstream users can pull from. We just haven't gotten around to set this up yet with either Travis CI or GitHub Actions (`help wanted`).

PR #204, opened 2020-03-03
https in CONTRIBUTING.md

koppor:
https everywhere

https://www.eff.org/https-everywhere

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,14 +4,14 @@ We welcome contributions to the Citation Style Language (CSL) locale files in th
 
 The CSL locale files provide the standard translations for automatic localization of CSL styles.
 As such, the CSL locale files should generally contain the most commonly used translations for a given locale.
-Less common translations can be [provided in the individual CSL styles](http://docs.citationstyles.org/en/stable/specification.html#locale) that require them, which will overwrite the CSL locale file translations.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -22,7 +22,7 @@ In addition, if you're interested in being credited, please add yourself to the
     <email>john.doe@citationstyles.org</email>
     <uri>https://citationstyles.org/</uri>
   </translator>
-  <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+  <rights license="https://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
```
adam3smith: This one won't work -- since we validate against the exact string here, we'd have to update all style files and there's really no reason to do that and thus make all last updated dates meaningless (especially since creativecommons.org automatically redirects to https anyway)

koppor: OK - I'll just withdraw the PR.

adam3smith: I'm actually happy to take the other two, especially since readthedocs doesn't enforce https; just revert this one.

rmzelle: @koppor, thanks, I made the changes in the freeform text at https://github.com/citation-style-language/locales/commit/d0ee4d13c942648c2c88629aeb01f03a17fac72c.

PR #206, opened 2020-04-01
Shino japanese

zaw-shinoto:
### Description

The older version should be completed, so if possible, translations were added, and in some cases, the older translation was changed where it seemed to be less common. Several terms seem to be not common in Japanese citations; they were translated where it felt natural. 

The translations needs testing and comparison with more Japanese journals. 

### Checklist

- [√ ] Check that you're listed as a `<translator>` in the `<info>` block at the beginning of the file.

### DIFF
--- a/locales-ja-JP.xml
+++ b/locales-ja-JP.xml
@@ -22,50 +22,50 @@
     <date-part name="day" suffix="日"/>
   </date>
   <terms>
-    <term name="accessed">参照</term>
+    <term name="accessed">アクセス</term>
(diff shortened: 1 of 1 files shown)


bwiernik:
@zaw-shinoto did you intend to close this?

zaw-shinoto:
sure.

zaw-shinoto:
I never got a reply and think this should be closed. Anyway, there are so many different solutions in different papers that it is better to solve the problem in the CSL rather than a locale.

PR #208, opened 2020-06-09
Update locales-hr-HR.xml

mkorlevic:
One of the Croatian translations of "et al." is "*i ostali*" but i haven't seen people using it. Instead, usually the term "*i suradnici*" (abbrev. "*i sur.*"; translates as "and collaborators") is used or rarely "*i drugi*" (abbrev. "*i dr.*"; translates as "and others") or even also rarely the term "et al." could be used as its origin is Latin. It would be great if all the terms could be stated, so that in the csl file one can specify the desired term.


bwiernik:
@mkorlevic Sorry on the long delay reviewing this!

Is the long form of "i suradnici" ever used, or only the abbreviation?

There actually is an `and-others` term, so we can specify "i dr." for that.

mkorlevic:
Hi!
I haven't seen people using the long form "_i suradnici_", however it would be great to also have this option.
The term `and-others` can be translated to "_i drugi_" (abbrev. "_i dr._").

bwiernik:
Thanks! Most citation styles do not specify to use the short form of "et al", so it's best to make the term the abbreviated form that is most widely used.

PR #210, opened 2020-11-26
migrate to GH actions

retorquere:
Updated for full dual-use between styles and locales.


retorquere:
Do not merge yet!

.github/workflows/merge.yaml
```diff
@@ -0,0 +1,98 @@
...
+        commit_message: Releasing ${{ steps.release.outputs.updated_files }} ${{ steps.release.outputs.deleted_files }}
+      if: github.event_name == 'push' && (steps.release.outputs.updated == 'true' || steps.release.outputs.deleted == 'true')
+
+    - uses: stefanzweifel/git-auto-commit-action@v4
+      with:
+        repository: 'release'
+        commit_message: ${{ github.event.inputs.commit_message }}
+      if: github.event_name == 'workflow_dispatch'
+
+    - name: ping distribution-updater
```
dstillman, line 94: Could you clarify that this is pinging Zotero more generally (styles page, Zotero client updates, etc.), not just the distribution-updater? I'd like to avoid a situation where the distribution-updater is removed in the future and someone forgets that this is also updating styles across the entire Zotero ecosystem.

.github/workflows/merge.yaml
```diff
@@ -0,0 +1,98 @@
...
+      if: github.repository == 'citation-style-language/styles'
+      run: |
+        curl -H "Authorization: $DISTRIBUTION_UPDATER_TOKEN" -F 'payload={"type":"pull_request","branch":"master","status":0,"commit":"'$GITHUB_SHA'"}' https://styles-update.zotero.org:8826/
+        curl -H "Authorization: $DISTRIBUTION_UPDATER_TOKEN" -F 'payload={"type":"pull_request","branch":"master","status":0,"commit":"'$GITHUB_SHA'"}' https://styles-update.zotero.org:8827/
```
dstillman, line 98: Shouldn't type be something other than `pull_request` here? These will just be skipped on our end.

.github/workflows/merge.yaml
rmzelle: > I'd like to avoid a situation where the distribution-updater is removed in the future and someone forgets that this is also updating styles across the entire Zotero ecosystem.

@dstillman, I think you told me before, but can you confirm that Zotero doesn't rely on https://github.com/citation-style-language/styles-distribution/ and https://github.com/citation-style-language/distribution-updater at all? I'm not sure how popular https://github.com/citation-style-language/styles-distribution/ is, but I think we'd want to deprecate it rather quickly and post a notification that people should now tune in at https://github.com/citation-style-language/styles/tree/v1.0.1 to get their vetted CSL 1.0.1 styles.

(cc @bwiernik, @adam3smith)

dstillman: We don't use styles-distribution, no. We just need the 8826 ping.

dstillman: When will >1.0.1 styles start being committed to `master`, though? If that's going to happen before Zotero supports them, we would need to make some changes on our end to make sure we weren't pushing out incompatible styles.

On that subject, @retorquere, it would be better to include the actual branch name for `branch`, not just hard-code "master".

rmzelle: > When will 1.1 styles start being committed to `master`, though?

@bwiernik might have an idea (I've not been very involved in the 1.0.2 and 1.1 release development). I did just propose a branch scheme for maintaining a 1.0.x dev branch. See https://github.com/citation-style-language/styles/pull/5126#issuecomment-734629766.

retorquere: > On that subject, @retorquere, it would be better to include the actual branch name for branch, not just hard-code "master".

No problem with that, but I thought `master` was the outcome of the earlier discussion. What is "the actual branch name" in this context? The `v1.0.1` release branch?

bwiernik: We will probably finish up the 1.1 schema and spec by January. From there, we'd probably want processors to update before we start to push 1.1 styles. Perhaps then would be a good time to notify clients and users to update their paths?

dstillman: > What is "the actual branch name" in this context?

Whatever commit the branch is on, the same as with the Travis requests.

If GH actions don't provide a variable for the branch, I think it would be the result of `git rev-parse --abbrev-ref HEAD`.

.github/workflows/merge.yaml
retorquere: Fixed. I glanced over the fact that it said `!=` rather than `==`.

.github/workflows/merge.yaml
retorquere: > Whatever branch the commit is on, the same as with the Travis requests.

which means "whatever branch *in the `release` directory is on", which in turn would translate to `v1.0.1` currently, correct? Because from the pov of the action, "the branch the commit is on" is always "master".

> If GH actions don't provide a variable for the branch, I think it would be the result of `git rev-parse --abbrev-ref HEAD`.

We have a var for the branch that the action is running on, and the `v1.0.1` branch we check out ourselves, so I can stuff that into a var for reuse.

rmzelle:
@retorquere, should this be brought up to date with https://github.com/citation-style-language/styles/blob/master/.github/workflows/merge.yaml?

retorquere:
Yes. But i could just have the styles workflow always commit the latest workflows to the locales workflow automatically. That way, this repo would be the one place to modify the workflows.

retorquere:
Uh, the other way around. Styles would be the source, this repo the target.

retorquere:
https://github.com/citation-style-language/styles/pull/5141 will copy the workflows into this repository when they're pushed to the styles repo.

.github/workflows/merge.yaml
adam3smith: Looks like @dstillman 's original concern here hasn't actually been addressed, right? so `ping zotero-updater` as a name for this?
And then using a variable for the branch name also isn't done yet, correct?

retorquere: If the locales and styles repos always have the same current-release version, https://github.com/citation-style-language/styles/pull/5141/files would take care of cross-applying changes to the workflows.

bwiernik: @retorquere I'm not sure how that answers either of Sebastian's questions?

retorquere: Since [this PR](https://github.com/citation-style-language/styles/pull/5141) would cause any change to the `styles` workflows to be copied over to the `locales` workflows, the line would automatically become "[ping Zotero servers](https://github.com/citation-style-language/styles/blob/56fde4bf291adb0e68482271310c644a5d5dd913/.github/workflows/merge.yaml#L102)".

adam3smith: got it, thanks

adam3smith:
@retorquere 
haven't looked at the output, but currently both of these runs fail:
https://github.com/citation-style-language/locales/actions/runs/398564446
https://github.com/citation-style-language/styles/actions/runs/398566280

retorquere:
The first fails because the `v1.0.1` branch does not exist.

The 2nd fails because it's checking out over https rather than ssh. It didn't do that when I tested it on my repos, looking into it.

retorquere:
OK the 2nd worked because I injected my creds. Fixing that.

PR #216, opened 2021-02-22
Add Hindi locale file

neha-sri:
### Description

This is the localization file for the Hindi language (India) which uses the Devanagari script. Devanagari alphabet is supported in UTF-8 so it should render accurately.

### Checklist

- [x] Validated the locale file at https://validator.citationstyles.org
- [x] Naming is as per recommended convention

### DIFF
--- a/locales-hi-IN.xml
+++ b/locales-hi-IN.xml
@@ -1,0 +1,361 @@
+<?xml version="1.0" encoding="utf-8"?>
+<locale xmlns="http://purl.org/net/xbiblio/csl" version="1.0" xml:lang="hi-IN">
+  <info>
+    <translator>
+      <name>Neha Srivastava</name>
+      <email>neha-srivastava@outlook.com</email>
+    </translator>
+    <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
+    <updated>2021-02-22T12:13:02+05:30</updated>
+  </info>
+  <style-options punctuation-in-quote="true"/>
+  <date form="text">
+    <date-part name="day" suffix=" "/>
+    <date-part name="month" suffix=", "/>
+    <date-part name="year"/>
+  </date>
+  <date form="numeric">
+    <date-part name="day" form="numeric-leading-zeros" suffix="/"/>
+    <date-part name="month" form="numeric-leading-zeros" suffix="/"/>
+    <date-part name="year"/>
+  </date>
+  <terms>
+    <term name="accessed">अंतिम उपयोग</term>
+    <term name="and">व</term>
+    <term name="and others">व अन्य</term>
+    <term name="anonymous">अज्ञात</term>
+    <term name="anonymous" form="short">अज्ञात</term>
+    <term name="at">पर</term>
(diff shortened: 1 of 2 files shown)


neha-sri:
Hello Maintainers, The check is failing because hindi is not listed in [locales.json ](https://github.com/citation-style-language/locales/blob/master/locales.json). I tried adding but I am not allowed to. Please take a look.

locales-hi-IN.xml
```diff
@@ -0,0 +1,361 @@
...
+    <term name="close-inner-quote">’</term>
+    <term name="page-range-delimiter">–</term>
+
+    <!-- ORDINALS -->
+    <term name="ordinal">वाँ</term>
+    <term name="ordinal-01"></term>
+    <term name="ordinal-02"></term>
+    <term name="ordinal-03"></term>
+    <term name="ordinal-04"></term>
+    <term name="ordinal-06"></term>
```
rmzelle, line 81: @neha-sri, can you explain how ordinal suffixes work in Hindi? (the "st" in "1st" and "nd" in "2nd")

Is "वाँ" always the suffix? If so, we should delete the other "ordinal-0x" terms here.

locales-hi-IN.xml
```diff
@@ -0,0 +1,361 @@
...
+    </term>
+
+    <!-- LONG ROLE FORMS -->
+    <term name="director">
+      <single>निर्देशक</single>
+      <multiple>directors</multiple>
+    </term>
+    <term name="editor">
+      <single>सम्पादक</single>
+      <multiple>editors</multiple>
```
rmzelle, line 262: @neha-sri, if necessary we can accept a partially translated CSL locale file, but is there a reason why you haven't translated some of the terms? (like the plural forms here)

locales-hi-IN.xml
neha-sri: So that part is complicated. Entire words change for 1st, 2nd, 3rd, 4th and 6th. वाँ doesnt work as an ordinal suffix for these numbers but then neither does any other suffix. I'll write a few so you'll get it. 

1st = पहला (péhlā) 
2nd = दूसरा (d̪oosrā)
3rd = तीसरा (t̪eesrā)
4th = चौथा (chaut̪hā)
5th = पाँचवाँ (pānchvān)
6th = छठा (chhat̪hhā)
7th = सातवाँ (sātvān)
8th = आठवाँ (āthvān)
9th = नवाँ (navān) 
10th = दसवाँ (dusvān) and so on.. 

So the ones with the वाँ can be abbreviated as 7वाँ but the others cannot be abbreviated as 1ला for example. Hence, this solution.

locales-hi-IN.xml
neha-sri: So I wanted to consult a few people and then submit the remaining words as a follow up change after trying this with a possible style. I can submit it on this PR also if thats preferable.

locales-hi-IN.xml
rmzelle: Okay, let's leave it as is, then.

locales-hi-IN.xml
rmzelle: I'll just merge it for now (it looks fine otherwise). Thanks, and we're looking forward to a more complete version!

PR #218, opened 2021-03-03
Update to Ruby 3.0.0

rmzelle:


rmzelle:
@retorquere, if you have a minute, any thoughts on how to best handle cases like this one? I'm trying to update the Ruby version used for GitHub Actions, but sheldon.yaml just skips the tests because no CSL locales were changed.

retorquere:
So, this is the tricky bit with PRs in GH actions. We want to be able to run tests like this which affect the tests themselves, but either 

1. only authorized people are able to start these tests, so they can be vetted, or
2. we move sheldon back out of GH actions so the exfiltration problem goes away.

The first one would be easiest to achieve with a `workflow_dispatch` action trigger. Upside: only people who have write access to the repo can start these, so the exfiltration problem. Downside: must be started manually.

The 2nd one removes the exfiltration problem entirely, and the the git repos must have the working state for Sheldon on Heroku. I liked having sheldon inlined, because I always forgot how some of the moving parts worked between travis and heroku, and there was no state to keep, but if it's going to be a roadblock, we may have to.

From a purely personal perspective, it introduces more moving parts *in Ruby*, and I've fallen out of love with how much magick ruby allows/needs/encourages, certainly for code touched very infrequently.

rmzelle:
> The first one would be easiest to achieve with a `workflow_dispatch` action trigger. Upside: only people who have write access to the repo can start these, so the exfiltration problem. Downside: must be started manually.

Being able to trigger the sheldon.yaml workflow manually seems like a great solution.

retorquere:
https://github.com/citation-style-language/styles/pull/5307

PR #225, opened 2021-04-20
Create locales-gl-GL.xml

mmgn897:
Translation to [Galician](https://en.wikipedia.org/wiki/Galician_language)


manuelmagan:
Could someone help me with this issue?

adam3smith:
Thanks -- ignore the failing test. 
I'm pretty sure GL as a country code doesn't exist. Would it be OK to use the standard gl-ES (see e.g. https://www.localeplanet.com/icu/gl-ES/index.html ) or is that politically problematic?

manuelmagan:
Thanks! Yes, gl-ES is okay.

manuelmagan:
Hello. Any news about this pull? Should I change something?

PR #230, opened 2021-06-18
add new terms for 1.0.2

denismaier:
Message edited by @bwiernik for documentation purposes.

New terms:

- General terms
  - "advance-online-publication"
  - "album"
  - "audio-recording"
  - "film"
  - "henceforth"
  - "loc-cit"
  - "no-place"
  - "no-publisher"
  - "on"
  - "op-cit"
  - "original-work-published"
  - "personal-communication"
  - "podcast"
  - "podcast-episode"
  - "preprint"
  - "radio-broadcast"
  - "radio-series"
  - "radio-series-episode"
  - "special-issue"
  - "special-section"
  - "television-broadcast"
  - "television-series"
  - "television-series-episode"
  - "video"
  - "working-paper"
  - "bce"
  - "ce"
  - "colon"
  - "comma"
  - "semicolon"

- Item type terms
  - "article"
  - "article-journal"
  - "article-magazine"
  - "article-newspaper"
  - "bill"
  - "book"
  - "broadcast"
  - "chapter"
  - "classic"
  - "collection"
  - "dataset"
  - "document"
  - "entry"
  - "entry-dictionary"
  - "entry-encyclopedia"
  - "event"
  - "graphic"
  - "hearing"
  - "interview"
  - "legal_case"
  - "legislation"
  - "manuscript"
  - "map"
  - "motion_picture"
  - "musical_score"
  - "pamphlet"
  - "paper-conference"
  - "patent"
  - "performance"
  - "periodical"
  - "personal_communication"
  - "post"
  - "post-weblog"
  - "regulation"
  - "report"
  - "review"
  - "review-book"
  - "software"
  - "song"
  - "speech"
  - "standard"
  - "thesis"
  - "treaty"
  - "webpage

- Locator terms
  - "act"
  - "appendix"
  - "article-locator"
  - "canon"
  - "elocation"
  - "equation"
  - "rule"
  - "scene"
  - "table"
  - "timestamp"
  - "title-locator"
  - renamed "sub verbo" to "sub-verbo"

- Role terms
  - "chair"
  - "compiler"
  - "contributor"
  - "curator"
  - "executive-producer"
  - "guest"
  - "host"
  - "narrator"
  - "organizer"
  - "performer"
  - "producer"
  - "script-writer"
  - "series-creator"


locales-en-US.xml
```diff
@@ -159,48 +132,48 @@
...
     <term name="canon">			 
-      <single></single>
-      <multiple></multiple>						 
+      <single>canon</single>
+      <multiple>canons</multiple>						 
     </term>
     <term name="elocation">			 
-      <single></single>
+      <single></single> <!-- ??? -->
       <multiple></multiple>
```
bwiernik:
```suggestion
      <single>location</single>
      <multiple>locations</multiple>						 
```

locales-en-US.xml
```diff
@@ -159,48 +132,48 @@
...
     </term>
     <term name="table">			 
-      <single></single>
-      <multiple></multiple>						 
+      <single>table</single>
+      <multiple>tables</multiple>						 
     </term>
     <term name="timestamp">			 
-      <single></single>
+      <single></single> <!-- ??? -->
```
bwiernik, line 171: @adam3smith I'm thinking leaving the timestamp terms blank by default? APA style doesn't label timestamps other than by the format (hh:mm:ss). I think that's generally the case with styles?

locales-en-US.xml
```diff
@@ -301,6 +525,19 @@
     </term>
 
     <!-- VERB ROLE FORMS -->
+    <term name="chair" form="verb"></term> <!-- ??? -->
```
bwiernik, line 528:
```suggestion
    <term name="chair" form="verb">chaired by</term>
```

locales-en-US.xml
```diff
@@ -275,6 +447,58 @@
     </term>
 
     <!-- SHORT ROLE FORMS -->
+    <term name="chair">
```
bwiernik, line 450: @adam3smith @bdarcus I could use a hand here. APA doesn't use short forms for any of these media roles. Do other fields?

locales-en-US.xml
```diff
@@ -301,6 +525,19 @@
     </term>
 
     <!-- VERB ROLE FORMS -->
+    <term name="chair" form="verb"></term> <!-- ??? -->
+    <term name="compiler" form="verb">compiled by</term>
+    <term name="contributor" form="verb">contributions by</term>
```
bwiernik, line 530:
```suggestion
    <term name="contributor" form="verb">with</term>
```

locales-en-US.xml
```diff
@@ -301,6 +525,19 @@
...
+    <term name="curator" form="verb">curated by</term>
+    <term name="executive-producer" form="verb">produced by</term>
```
bwiernik, line 532:
```suggestion
    <term name="executive-producer" form="verb">executive produced by</term>
```

locales-en-US.xml
```diff
@@ -301,6 +525,19 @@
...
+    <term name="guest" form="verb"></term> <!-- ??? -->
```
bwiernik, line 533:
```suggestion
    <term name="guest" form="verb">with guest</term>
```

@adam3smith What do you think? Or just `with`?

locales-en-US.xml
```diff
@@ -301,6 +525,19 @@
...
+    <term name="host" form="verb">hosted by</term> <!-- ??? -->
```
bwiernik, line 534:
```suggestion
    <term name="host" form="verb">hosted by</term>
```

locales-en-US.xml
```diff
@@ -313,6 +550,19 @@
     <term name="editortranslator" form="verb">edited &amp; translated by</term>
 
     <!-- SHORT VERB ROLE FORMS -->
+    <term name="chair" form="verb-short"></term> <!-- ??? -->
+    <term name="compiler" form="verb-short">compiled by</term>
+    <term name="contributor" form="verb-short">contributions by</term>
```
bwiernik, line 555: @adam3smith @rmzelle @fbennett Can you remind me, if a form is included here, but blank, that is treated as "missing" and falls back to the long form, correct? So many of these could be reduced to blank if the default should be fallback?

```suggestion
    <term name="contributor" form="verb-short">with</term>
```

bwiernik (changes requested):
@adam3smith Some questions for you

locales-en-US.xml
```diff
@@ -25,6 +25,33 @@
     <date-part name="year"/>
   </date>
   <terms>
+    <term name="advance-online-publication">advance online publication</term>
+    <term name="album">album</term>
+    <term name="audio-recording">audio recording</term>
+    <term name="film">film</term>
+    <term name="henceforth">henceforth</term>
+    <term name="loc-cit">loc. cit.</term> <!-- like ibid., the abbreviated form is the regular form  -->
+    <term name="no-place">no place</term>
```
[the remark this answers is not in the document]
adam3smith: I'd like to see evidence on how universal this is used in Latin. Chicago Manual uses n.p. as does MLA 7 (I think they got rid of this entirely in 8+).

locales-en-US.xml
```diff
@@ -25,6 +25,33 @@
...
+    <term name="no-place" form="short">n. p.</term>
+    <term name="no-publisher">no publisher</term>
```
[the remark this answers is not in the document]
adam3smith: Ditto here: MLA also uses n.p. here (yes, same acronym). Chicago doesn't know this case.

locales-en-US.xml
adam3smith: No, it only falls back to long when the term is missing, not when it's blank. 
A blank term means nothing is being rendered (as you do for the timestamp); that extends to short forms as well (just tested with citeproc-js; not sure if the behavior is in the specs -- I don't think so -- but I think it's correct).

locales-en-US.xml
adam3smith: I think this works, yes

locales-en-US.xml
```diff
@@ -25,6 +25,33 @@
...
+    <term name="no-publisher" form="short"></term> <!-- ??? -->
+    <term name="on">on</term>
+    <term name="op-cit">op. cit.</term> <!-- like ibid., the abbreviated form is the regular form  -->
+    <term name="original-work-published"></term> <!-- ??? -->
```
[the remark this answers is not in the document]
adam3smith: or "originally published"?

locales-en-US.xml
adam3smith: agree.

locales-en-US.xml
adam3smith: I think APA may be the only style that uses chair in citations. I'd just set the short form to "chair" as well (I've seen chr. as an acronym, but I don't think it's particularly common or useful).

adam3smith:
I think I answered all queries & added some comments on @bwiernik 's suggested edits (haven't looked at the full PR, though)

locales-en-US.xml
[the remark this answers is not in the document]
bwiernik: APA uses "original work published". Anything specific from Chicago or MLA?

locales-en-US.xml
```diff
@@ -99,6 +131,50 @@
...
+    <term name="scene">			 
+      <single>scene</single>
+      <multiple>scenes</multiple>						 
+    </term>
+    <term name="table">			 
+      <single>table</single>
+      <multiple>tables</multiple>						 
+    </term>
+    <term name="timestamp">			 
+      <single></single> <!-- ??? -->
```
bwiernik, line 171:
```suggestion
      <single></single> <!-- generally blank -->
```

locales-en-US.xml
```diff
@@ -99,6 +131,50 @@
...
+    </term>
+    <term name="scene">			 
+      <single>scene</single>
+      <multiple>scenes</multiple>						 
+    </term>
+    <term name="table">			 
+      <single>table</single>
+      <multiple>tables</multiple>						 
+    </term>
+    <term name="timestamp">
```
bwiernik, line 170:
```suggestion
    <term name="timestamp"> <!-- generally blank -->
```

locales-en-US.xml
bwiernik: @adam3smith Could you help out with compiler/curator/producer/etc. below?

locales-en-US.xml
bwiernik: @adam3smith Thanks. In that case, could you help out with the short forms for the verbs here as well? I'm not sure how they would generally be abbreviated.

locales-en-US.xml
```diff
@@ -169,6 +245,50 @@
...
+    </term>
+    <term name="scene">			 
+      <single></single>
+      <multiple></multiple>						 
+    </term>
+    <term name="table">			 
+      <single></single>
+      <multiple></multiple>						 
+    </term>
+    <term name="timestamp">
```
bwiernik, line 284:
```suggestion
    <term name="timestamp"> <!-- generally blank -->
```

locales-en-US.xml
```diff
@@ -169,6 +245,50 @@
...
+      <single></single>
+      <multiple></multiple>						 
+    </term>
+    <term name="canon">			 
+      <single></single>
+      <multiple></multiple>						 
+    </term>
+    <term name="elocation">			 
+      <single></single>
+      <multiple></multiple>
```
bwiernik, line 266:
```suggestion
      <single>loc.</single>
      <multiple>locs.</multiple>
```

locales-en-US.xml
```diff
@@ -169,6 +245,50 @@
...
+      <single></single>
+      <multiple></multiple>						 
+    </term>
+    <term name="appendix">			 
+      <single></single>
+      <multiple></multiple>						 
+    </term>
+    <term name="article-locator">			 
+      <single></single>
+      <multiple></multiple>
```
bwiernik, line 258:
```suggestion
      <single>art.</single>
      <multiple>arts.</multiple>
```

locales-en-US.xml
```diff
@@ -169,6 +245,50 @@
...
+      <single></single>
+      <multiple></multiple>						 
+    </term>
+    <term name="elocation">			 
+      <single></single>
+      <multiple></multiple>						 
+    </term>
+    <term name="equation">			 
+      <single></single>
+      <multiple></multiple>
```
bwiernik, line 270:
```suggestion
      <single>eq.</single>
      <multiple>eqs.</multiple>
```

locales-en-US.xml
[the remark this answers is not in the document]
bwiernik: If Chicago and MLA do in English, I'm fine with the baseline being that way.

locales-en-US.xml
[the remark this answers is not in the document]
adam3smith: Chicago is all over the place. There's "first published" for reprints (14.114) and "originally published" for translations (14.99). 
MLA has "originally published" (https://style.mla.org/magazine-essay-on-web-site/ )

locales-en-US.xml
```diff
@@ -313,6 +550,19 @@
     <term name="editortranslator" form="verb">edited &amp; translated by</term>
 
     <!-- SHORT VERB ROLE FORMS -->
+    <term name="chair" form="verb-short"></term> <!-- ??? -->
+    <term name="compiler" form="verb-short">compiled by</term>
+    <term name="contributor" form="verb-short">contributions by</term>
+    <term name="curator" form="verb-short">curated by</term>
+    <term name="executive-producer" form="verb-short">produced by</term>
+    <term name="guest" form="verb-short"></term> <!-- ??? -->
```
bwiernik, line 558:
```suggestion
    <term name="guest" form="verb-short">with guest</term>
```

locales-en-US.xml
```diff
@@ -313,6 +550,19 @@
     <term name="editortranslator" form="verb">edited &amp; translated by</term>
 
     <!-- SHORT VERB ROLE FORMS -->
+    <term name="chair" form="verb-short"></term> <!-- ??? -->
+    <term name="compiler" form="verb-short">compiled by</term>
```
adam3smith, line 554: I don't think we'll find an authoritative source here. I don't think any of these are in MLA or Chicago and thus likely also missing from other works. We could just take best guesses? (e.g. "comp. by" is likely right, and I'd abbreviate with as "w.", etc.) Any other ideas?

locales-en-US.xml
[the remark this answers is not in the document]
denismaier: Good, so we keep "no place"

locales-en-US.xml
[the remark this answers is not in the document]
denismaier: Whatever we choose here should this term include an "as"? As in "originally published **as**"?

locales-en-US.xml
```diff
@@ -99,6 +131,50 @@
...
+    </term>
+    <term name="article-locator">			 
+      <single>article</single>
+      <multiple>articles</multiple>						 
+    </term>
+    <term name="canon">			 
+      <single>canon</single>
+      <multiple>canons</multiple>						 
+    </term>
+    <term name="elocation">
```
denismaier, line 150: @bwiernik you've added a short form for this. Shouldn't we then also add a long form for this?
(OTOH, I'm a bit unsure about this having a term at all. Can't this just stay empty, just as timestamps?)

locales-en-US.xml
```diff
@@ -169,6 +245,50 @@
...
+    </term>
+    <term name="article-locator">			 
+      <single>art.</single>
+      <multiple>arts.</multiple>
+    </term>
+    <term name="canon">			 
+      <single></single>
+      <multiple></multiple>						 
+    </term>
+    <term name="elocation">
```
denismaier, line 264: Can't this just stay empty, just as timestamps?

locales-en-US.xml
bwiernik: @adam3smith What do you think? I've mostly seen kindle editions, etc. cited with the term "location" when these are cited (versus page numbers)

locales-en-US.xml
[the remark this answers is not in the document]
bwiernik: No, for example, the APA phrasing is "Original work published 1962". Though, this is a good point that we might need an `as` term.

locales-en-US.xml
```diff
@@ -313,6 +550,19 @@
...
+    <term name="curator" form="verb-short">curated by</term>
+    <term name="executive-producer" form="verb-short">produced by</term>
+    <term name="guest" form="verb-short">with guest</term>
+    <term name="host" form="verb-short">hosted by</term> <!-- ??? -->
+    <term name="narrator" form="verb-short">narrated by</term>
+    <term name="organizer" form="verb-short">organized by</term>
+    <term name="performer" form="verb-short">performed by</term>
+    <term name="producer" form="verb-short">produced by</term>
+    <term name="script-writer" form="verb-short">written by</term>
+    <term name="series-creator" form="verb-short">created by</term>
```
bwiernik, line 657:
```suggestion
    <term name="contributor" form="verb-short">w.</term>
    <term name="curator" form="verb-short">cur. by</term>
    <term name="executive-producer" form="verb-short">exec. prod. by</term>
    <term name="guest" form="verb-short">w. guest</term>
    <term name="host" form="verb-short">hosted by</term>
    <term name="narrator" form="verb-short">narr. by</term>
    <term name="organizer" form="verb-short">org. by</term>
    <term name="performer" form="verb-short">perf. by</term>
    <term name="producer" form="verb-short">prod. by</term>
    <term name="script-writer" form="verb-short">writ. by</term>
    <term name="series-creator" form="verb-short">cre. by</term>
```

locales-en-US.xml
bwiernik, line 646:
```suggestion
    <term name="chair" form="verb-short">chaired by</term>
    <term name="compiler" form="verb-short">comp. by</term>
```

locales-en-US.xml
```diff
@@ -275,6 +447,58 @@
...
+      <multiple></multiple>
+    </term>
+    <term name="script-writer">
+      <single></single>
+      <multiple></multiple>
+    </term>
+    <term name="series-creator">
+      <single></single>
+      <multiple></multiple>
+    </term>
```
bwiernik, line 593:
```suggestion
      <single>chair</single>
      <multiple>chairs</multiple>
    </term>
    <term name="compiler">
      <single>comp.</single>
      <multiple>comps.</multiple>
    </term>
    <term name="contributor">
      <single>contrib.</single>
      <multiple>contribs.</multiple>
    </term>
    <term name="curator">
      <single>cur.</single>
      <multiple>curs.</multiple>
    </term>
    <term name="executive-producer">
      <single>exec. prod.</single>
      <multiple>exec. prods.</multiple>
    </term>
    <term name="guest">
      <single>guest</single>
      <multiple>guests</multiple>
    </term>
    <term name="host">
      <single>host</single>
      <multiple>hosts</multiple>
    </term>
    <term name="narrator">
      <single>narr.</single>
      <multiple>narrs.</multiple>
    </term>
    <term name="organizer">
      <single>org.</single>
      <multiple>orgs.</multiple>
    </term>
    <term name="performer">
      <single>perf.</single>
      <multiple>perfs.</multiple>
    </term>
    <term name="producer">
      <single>prod.</single>
      <multiple>prods.</multiple>
    </term>
    <term name="script-writer">
      <single>writ.</single>
      <multiple>writs.</multiple>
    </term>
    <term name="series-creator">
      <single>cre.</single>
      <multiple>cres.</multiple>
    </term>
```

locales-en-US.xml
```diff
@@ -249,6 +369,58 @@
...
     <!-- LONG ROLE FORMS -->
+    <term name="chair">
+      <single>chair</single>
+      <multiple>chairs</multiple>
+    </term>
+    <term name="compiler">
+      <single>compiler</single>
+      <multiple>compilers</multiple>
+    </term>
+    <term name="contributor">
```
bwiernik, line 472: @adam3smith What do you think? Should `contributor` just have the `with` verb form and no regular form, like with `author`?

bwiernik:
@denismaier Can you add the term stubs for each item type (report, book, broadcast, etc.)?

locales-en-US.xml
```diff
@@ -169,6 +245,50 @@
     </term>
 
     <!-- SHORT LOCATOR FORMS -->
+    <term name="act">			 
+      <single>a.</single> <!-- according to acronyms.com, but is this really common?-->
+      <multiple></multiple>
```
bwiernik, line 342:
```suggestion
      <single>act</single>
      <multiple>acts</multiple>						 
```

Indigobook and SBL don't have an abbreviation for act, so let's go without unless there is some complaint

locales-en-US.xml
```diff
@@ -169,6 +245,50 @@
...
+      <multiple>apps.</multiple>						 
+    </term>
+    <term name="article-locator">			 
+      <single>art.</single>
+      <multiple>arts.</multiple>
+    </term>
+    <term name="canon">			 
+      <single></single>
+      <multiple></multiple>						 
+    </term>
```
bwiernik, line 355:
```suggestion
    <term name="canon">			 
      <single>canon</single>
      <multiple>canons</multiple>						 
    </term>
```

locales-en-US.xml
```diff
@@ -169,6 +245,50 @@
...
+      <single>loc.</single>
+      <multiple>locs.</multiple>
+    </term>
+    <term name="equation">			 
+      <single>eq.</single>
+      <multiple>eqs.</multiple>
+    </term>
+    <term name="rule">			 
+      <single></single>
+      <multiple></multiple>
```
bwiernik, line 366:
```suggestion
      <single>r.</single>
      <multiple>rr.</multiple>						 
```

denismaier:
> @denismaier Can you add the term stubs for each item type (report, book, broadcast, etc.)?

done.

locales-en-US.xml
```diff
@@ -169,6 +245,50 @@
...
+      <single>tab.</single>
+      <multiple>tabs.</multiple>						 
+    </term>
+    <term name="timestamp"> <!-- generally blank -->
+      <single></single>
+      <multiple></multiple>						 
+    </term>
+    <term name="title-locator">			 
+      <single></single>
+      <multiple></multiple>
```
bwiernik, line 382:
```suggestion
      <single>tit.</single>
      <multiple>tits.</multiple>
```

locales-en-US.xml
```diff
@@ -249,6 +369,58 @@
...
+      <single>performer</single>
+      <multiple>performers</multiple>
+    </term>
+    <term name="producer">
+      <single>producer</single>
+      <multiple>producers</multiple>
+    </term>
+    <term name="script-writer"> <!-- shouldn't that be scriptwriter?-->
+      <single>scriptwriter</single>
+      <multiple>scriptwriters</multiple>
```
bwiernik, line 510:
```suggestion
      <single>writer</single>
      <multiple>writers</multiple>
```

> `<!-- shouldn't that be scriptwriter?-->`

It was originally just "writer". The "script" was added during review to reduce confusion. I don't have a problem either way, but the variable name that's incorporated into pandoc, etc. at this point is `script-writer`, so we should probably keep it as is.

@adam3smith

locales-en-US.xml
bwiernik: What do you think of the below?

locales-en-US.xml
denismaier: Ok. Should we just delete the short term form? (Will it fall back to the long form in the absence of a short form?)

locales-en-US.xml
denismaier: same here.

locales-en-US.xml
denismaier: Or should we short form entries that repeat the long form?

locales-en-US.xml
```diff
@@ -169,6 +337,50 @@
...
+      <single>r.</single>
+      <multiple>rr.</multiple>						 
+    </term>
+    <term name="scene">			 
+      <single>sc.</single>
+      <multiple>scs.</multiple>						 
+    </term>
+    <term name="table">			 
+      <single>tab.</single>
+      <multiple>tabs.</multiple>
```
bwiernik, line 374:
```suggestion
    <term name="table">			 
      <single>tbl.</single>
      <multiple>tbls.</multiple>						 
```

locales-en-US.xml
denismaier: Pandoc etc can still be changed easily so that's not that important.

locales-en-US.xml
denismaier: Ok. In that case I'd say we add "location/locations" as the long form. (We shouldn't be to strict here anyway as styles can always override the defaults. And if something doesn't make sense people can always complain and we make reasonable changes later. But for this, folks need a chance to test use this.)

locales-en-US.xml
```diff
@@ -66,16 +93,113 @@
...
+    <term name="document" form="short"></term>
+    <term name="entry">entry</term>
+    <term name="entry" form="short"></term>
+    <term name="entry-dictionary">dictionary entry</term>
+    <term name="entry-dictionary" form="short"></term>
+    <term name="entry-encyclopedia">encyclopedia entry</term>
+    <term name="entry-encyclopedia" form="short"></term>
+    <term name="event">event</term>
+    <term name="event" form="short"></term>
+    <term name="figure">figure</term> <!-- there is an existing figure term right? -->
```
denismaier, line 129: @bwiernik Yes, under locator we already have a `figure` term. In other similar cases we've renamed the locator term to `X-locator`, as in `article` vs. `article-locator` or `title` vs. `title-locator`.

locales-en-US.xml
denismaier: @bwiernik I'd leave it in for now. (I agree that this will be mostly be used like "Jane Smith, with John Doe. *A very important book* ...", but it won't hurt to have that in here as well. Ok?

denismaier:
We should be almost ready, but one general question: How do we deal with short forms being identical to the long forms? Omit the short form? Or keep both?

locales-en-US.xml
```diff
@@ -66,16 +93,113 @@
...
+    <term name="speech"></term>
+    <term name="speech" form="short"></term>
+    <term name="standard"></term>
+    <term name="standard" form="short"></term>
+    <term name="thesis"></term>
+    <term name="thesis" form="short"></term>
+    <term name="treaty"></term>
+    <term name="treaty" form="short"></term>
+    <term name="webpage"></term>
+    <term name="webpage" form="short"></term>
```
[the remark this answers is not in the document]
denismaier: > For a default term for `article`, I'm thinking "preprint" for now?

We also have a dedicated term `preprint`.

denismaier:
More questions:
- should we now add an `as` term?
- should there be different `original-work-published` terms for reprints and translations?

bwiernik:
I think reprinted/translated would be addressed as part of the related items structure, so I'm inclined to not add those until a better mechanism exists.

locales-en-US.xml
denismaier: @bwiernik What do you think about this? Would renaming the existing figure term to `figure-locator` make sense?

denismaier:
@bwiernik What do you think is still missing? 

Also, repeating my question from above:

> We should be almost ready, but one general question: How do we deal with short forms being identical to the long forms? Omit the short form? Or keep both?

bwiernik:
I think the fall back for short form terms is long form? Can you check that?

denismaier:
> I think the fall back for short form terms is long form? Can you check that?

I'm not 100% sure about the fallback logic, but looking at the [verb role forms in the German locale](https://github.com/citation-style-language/locales/blob/ec6d62a9e7c049001a55086a495022520989441c/locales-de-CH.xml#L264) I'd say that this must be the case. (Certain short forms aren't defined.) Also: A test with Chicago Note Bibliography confirms that.

It might neverthless be a good idea to just use the long form even for the short form in those cases as this might help with copying the new terms to other locales.

@adam3smith @rmzelle Thoughts?

adam3smith:
I could go either way on this, but I think would softly lean against adding short forms across the board: Since we start with English for anything unknown, that creates a lot more work during translation. But don't feel strongly, so if others disagree or have good arguments to include them, I'm happy to be overruled.

bwiernik:
I lean toward not including the short forms. The fallback to long forms is part of the specification, so I think it's okay to make the locales leaner by default (and avoid unintended collisions with English).

denismaier:
Ok. Looks like that's the way to go. I'll update the PR.

denismaier:
@bwiernik I've updated the PR. (Don't know why the checks fail... Can you see why?)
Anyway, I've corrected some mistakes, removed redundant short forms and refactored the item type terms into two distinct lists, one for long forms, the other for short forms---just as we do with the other terms. Could you please add the abbreviations, and delete those where there's no abbreviated form?

locales-en-US.xml
denismaier: @bwiernik What do you think about this? Should we rename it to `figure-locator`?

locales-en-US.xml
bwiernik: sure.

locales-en-US.xml
bwiernik: We need terms for both `figure` and `figure-locator` (and `article` and `article-locator`)

denismaier: Ok, I've renamed the locator term to `figure-locator`. We have both terms for the locator and for the item type now.

adam3smith:
I'm pretty sure the automated test would validate against 1.0.1, which obviously fails

denismaier:
> I'm pretty sure the automated test would validate against 1.0.1, which obviously fails

(Just checked: this is what the code says: https://github.com/citation-style-language/locales/blob/d5ee85de8e74d4109509014758b6f496a968ff03/spec/locales_spec.rb#L5)

Ok, so failing tests are not necessarily a bad sign in that case.

bwiernik:
Okay, these all validate against the 1.0.2 schema

zepinglee:
IMHO the trailing spaces (e.g., [this line](https://github.com/denismaier/locales/blob/0c3396c5a7c1d5d3aa01e43bf0a58d0f9b45c069/locales-en-US.xml#L198)) are supposed to be removed. They may cause extra diff lines when modified.

bdarcus:
> * added some terms

FWIW, would probably help to itemize these, so they show up in searches in the future.

I, for example, couldn't find an existing issue or PR that added "figure-locator", so seemed like a mistake.

PR #231, opened 2021-06-23
Update of PL "accessed" translation 

Pendzoncymisio:
### Description
Changing the translation of "accessed" to be correct. Currently translation of "accessed" means something like "shared". In update I am proposing translation that is widely used in Polish citations.

### Checklist
- [ X ] Check that you're listed as a `<translator>` in the `<info>` block at the beginning of the file.

### DIFF
--- a/locales-pl-PL.xml
+++ b/locales-pl-PL.xml
@@ -25,7 +25,7 @@
     <date-part name="year"/>
   </date>
   <terms>
-    <term name="accessed">dostęp:</term>
+    <term name="accessed">dostęp</term>
     <term name="and">i</term>
     <term name="and others">i inni</term>
     <term name="anonymous">anonim</term>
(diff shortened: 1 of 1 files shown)


bwiernik:
Should the `:` be included there? The term should just be the word used for "accessed", not any punctuation, which will be determined by the style

Pendzoncymisio:
You are right, this colon should be specified by the style, fixed it.

bwiernik:
@adam3smith @rmzelle Any special steps needed before merging changes to locales?

adam3smith:
@bwiernik -- where possible (unless we know the language very well) we try to get one of the previous contributors to take a quick look when terms are changed, but the last substantive changes here are 8 years ago and the term in question >10, so I'd just take this.

bwiernik:
got it thanks!

I also checked Google scholar and found many examples of this term.

PR #242, opened 2021-11-13
Update locales-zh-CN.xml

zepinglee:
### Description

1. Correct the short form of `edition` term.
The short form of `edition` term should be "版" not "本". See <https://translate.google.com/?sl=en&tl=zh-CN&text=edition&op=translate>.

2. Correct quotation marks.
The quotation marks in Simplified Chinese should be U+201C, U+201C, U+2018, and U+2019 as specified in the China national standard [GB/T 15834—2011 General rules for punctuation](http://std.samr.gov.cn/gb/search/gbDetailed?id=71F772D7FE25D3A7E05397BE0A0AB82A) (in Chinese). See also <https://en.wikipedia.org/wiki/Quotation_mark#Summary_table>. The original fullwidth angled brackets "《》〈〉" are actually used for indicating book titles in Chinese.

3. Change `page-range-delimiter` to fullwidth tilde "～" (U+FF5E).
The fullwidth tilde "～" (U+FF5E) is used as the delimiter of numeric range in general Chinese language, which is also specified in GB/T 15834—2011. See also <https://en.wikipedia.org/wiki/Tilde#Range>. As an exception, the `page-range-delimiter` is hyphen in the GB/T 7714—2015 China national bibliography style. Thus the `china-national-standard-gb-t-7714-2015-*.csl` from <https://github.com/citation-style-language/styles> need to be updated after this PR is merged.

BTW, there are a lot of trailing spaces in the file which may need to be cleaned.

### Checklist

### DIFF
--- a/locales-zh-CN.xml
+++ b/locales-zh-CN.xml
@@ -10,19 +10,22 @@
     <translator>
       <name>Heromyth</name>
     </translator>
+    <translator>
+      <name>Zeping Lee</name>
+    </translator>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
     <updated>2014-05-15T23:31:02+00:00</updated>
   </info>
   <style-options punctuation-in-quote="false"/>
   <date form="text">
-    <date-part name="year" suffix="年"/>
-    <date-part name="month" form="numeric" suffix="月"/>
-    <date-part name="day" suffix="日"/>
+    <date-part name="year" suffix="年" range-delimiter="&#8212;"/>
+    <date-part name="month" form="numeric" suffix="月" range-delimiter="&#8212;"/>
+    <date-part name="day" suffix="日" range-delimiter="&#8212;"/>
   </date>
   <date form="numeric">
-    <date-part name="year"/>
-    <date-part name="month" form="numeric-leading-zeros" prefix="/"/>
-    <date-part name="day" form="numeric-leading-zeros" prefix="/"/>
+    <date-part name="year" range-delimiter="/"/>
+    <date-part name="month" form="numeric-leading-zeros" prefix="-" range-delimiter="/"/>
+    <date-part name="day" form="numeric-leading-zeros" prefix="-" range-delimiter="/"/>
   </date>
   <terms>
     <term name="advance-online-publication">advance online publication</term>
@@ -64,7 +67,7 @@
     <term name="circa" form="short">约</term>
     <term name="cited">见引于</term>
     <term name="edition">版本</term>
-    <term name="edition" form="short">本</term>
+    <term name="edition" form="short">版</term>
     <term name="et-al">等</term>
     <term name="forthcoming">即将出版</term>
     <term name="from">从</term>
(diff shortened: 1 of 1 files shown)


zepinglee:
Additional commit:

4. The official date form format in China is YYYY-MM-DD as specified in [GB/T 7408—2005 Data elements and interchange formats—Representation of dates and times](http://std.samr.gov.cn/gb/search/gbDetailed?id=71F772D77FD1D3A7E05397BE0A0AB82A) (in Chinese) which is based on ISO 8601. See also entry China in <https://en.wikipedia.org/wiki/Date_format_by_country#Listing>. This format is also used in the GB/T 7714—2015 China national bibliography style.

bwiernik:
With respect to the quote marks, in citation styles, quote marks are mostly used for book and article titles, so the angle brackets would be correct, no?

zepinglee:
> With respect to the quote marks, in citation styles, quote marks are mostly used for book and article titles, so the angle brackets would be correct, no?

@bwiernik It may cause problem for a title with quoted text such as `Foo "bar" baz`. CSL converts those quotation marks to angle brackets and the result is `Foo 《bar》 baz`. But it doesn't make sense since the `bar` itself is not an article or book title.

bwiernik:
@adam3smith Not sure how to handle this here because CSL doesn't distinguish between types of quotes like that. 

The context where angle brackets would be used is by far the most common case, so even if CSL can't currently handle all cases correctly, reverting the outer brackets to angle brackets should be done.

zepinglee:
It sounds reasonable if the `*-quote` terms are specific to article titles. I'll revert that change.

I also checked the following example.
```xml
<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" default-locale="zh-CN">
  <info>
    <id />
    <title />
    <updated>2009-08-10T04:49:00+09:00</updated>
  </info>
  <locale xml:lang="zh-CN">
    <terms>
      <term name="open-quote">《</term>
      <term name="close-quote">》</term>
      <term name="open-inner-quote">〈</term>
      <term name="close-inner-quote">〉</term>
    </terms>
  </locale>
  <citation>
    <layout>
      <text value="Foo &#34;bar&#34; baz" quotes="true"/>
      <text value="Foo “bar” baz" quotes="true"/>
      <text value="甲&#34;乙&#34;丙" quotes="true"/>
      <text value="甲“乙”丙" quotes="true"/>
    </layout>
  </citation>
</style>
```
`Citeproc-js` gives `《Foo 〈bar〉 baz》《Foo “bar” baz》《甲"乙"丙》《甲“乙”丙》` thus it is ok for unicode quotation marks as well as pure Chinese.

zepinglee:
The reverting is done. I've also added changes about the date range.

- Em dash is used in text form of date. This is specified in *GB/T 15834—2011 General rules for punctuation* (in Chinese) where an example "2011 年 2 月 3 日—10 日" is directly given in section 4.13..3.2.
- The date range-delimiter in numeric form is changed "/" as specified in *GB/T 7408—2005 Data elements and interchange formats—Representation of dates and times* (in Chinese) which is based on ISO 8601.

Besides, can I squash these changes to one commit before merging?

zepinglee:
The changes are rebased on master branch (4a551a8) to solve the conflicts.

I also added zh-CN translations for new terms in v1.0.2 and they are reviewed by @redleafnew (https://github.com/zepinglee/csl-locales/pull/1 in Chinese).

PR #244, opened 2021-12-23
Rename "figure" and "sub verbo" for 1.0.2

bwiernik:
Rename these terms. Should be the only change needed for validation on CSL 1.0.2

### DIFF
--- a/locales-af-ZA.xml
+++ b/locales-af-ZA.xml
@@ -102,7 +102,7 @@
       <single>column</single>
       <multiple>columns</multiple>
     </term>
-    <term name="figure">
+    <term name="figure-locator">
       <single>figure</single>
       <multiple>figures</multiple>
     </term>
@@ -146,7 +146,7 @@
       <single>section</single>
       <multiple>sections</multiple>
     </term>
-    <term name="sub verbo">
+    <term name="sub-verbo">
       <single>sub verbo</single>
       <multiple>sub verbis</multiple>
     </term>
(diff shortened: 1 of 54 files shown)


bdarcus:
There are a bunch of errors like this:

```console
locales-af-ZA.xml:105:4: error: attribute ^name with invalid value "figure-locator"
```

adam3smith:
Huh, that suggests something wrong with the schema? We added that, no?

bwiernik:
Is it validating against 1.0.1 not 1.0.2?

bwiernik:
It's definitely `sub-verbo` in the terms schema : https://github.com/citation-style-language/schema/blob/master/schemas/styles/csl-terms.rnc and https://github.com/citation-style-language/schema/blob/v1.0.2/schemas/styles/csl-terms.rnc

@bdarcus Did you somehow validate against 1.0.1 instead of 1.0.2?

bdarcus:
> Huh, that suggests something wrong with the schema? We added that, no?

"sub-verbo" is correct, but "figure-locator"?

Also, I can't find any record of "figure-locator", and it looks odd to me alongside the other terms.

> @bdarcus Did you somehow validate against 1.0.1 instead of 1.0.2?

I was validating using the master branch. Shouldn't that be correct? It's not, but ...

bwiernik:
figure-locator because there is also a term for the item type figure

bwiernik:
What is the validation command you ran @bdarcus ?

adam3smith:
I was only tangentially involved in the discussion, but this sounds right to me. We also don't have page and page-locator

bwiernik:
Closing this in favor of https://github.com/citation-style-language/locales/pull/230

denismaier:
Yes. I also think we don't necessarily need figure-locator. I've included it solely for the sake of consistency.

PR #245, opened 2022-01-01
Update Japanese (ja-JP) Localization

cmplstofB:
### Description

The following documents were used as references for the Japanese translation.

* SIST 02:2007「参照文献の書き方」
* JIS X 0807:1999「電子文献の引用法」

As a result, the existing translation has been slightly modified.
For example, "retrieved" should be translated as "入手元" (cf: JIS X 0807:1999, §6.12).

Some words (e.g. `pp.` the short form for the page ranges) are still in English, but they have been translated keeping their form, not untranslated.

### DIFF
--- a/locales-ja-JP.xml
+++ b/locales-ja-JP.xml
@@ -3,9 +3,15 @@
   <info>
     <translator>
       <name>Shoji Takahashi</name>
+      <name>cmplstofB</name>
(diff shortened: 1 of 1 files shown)


cmplstofB:
Sorry, I pulled invaild XML file. Close.

PR #246, opened 2022-01-01
 Update Japanese (ja-JP) Localization [fix for #245]

cmplstofB:
The following documents were used as references for the Japanese translation.

* SIST 02:2007「参照文献の書き方」
* JIS X 0807:1999「電子文献の引用法」

As a result, the existing translation has been slightly modified.
For example, "retrieved" should be translated as "入手元" (cf: JIS X 0807:1999, §6.12).

Some words (e.g. `pp.` the short form for the page ranges) are still in English, but they have been translated keeping their form, not untranslated.

---

This issue is fix for #245

### DIFF
--- a/locales-ja-JP.xml
+++ b/locales-ja-JP.xml
@@ -3,6 +3,8 @@
   <info>
     <translator>
       <name>Shoji Takahashi</name>
+    </translator>
+    <translator>
       <name>cmplstofB</name>
     </translator>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thank you @cmplstofB!

There are some new terms in CSL 1.0.2 -- if you have time, it would be awesome if you could translate those too!

cmplstofB:
@bwiernik 

I updated localization file for CSL 1.0.2 in #292.

PR #254, opened 2022-03-12
en-GB & en-US: Add missing form attribute to terms

ellen364:
Add missing `form="short"` to terms in the "short locator forms" section of locales-en-GB.xml and locales-en-US.xml.

The fix is small but could surprise downstream users. When `form="short"` is missing, some long locator forms (e.g. appendix) get redefined. Depending on how downstream users parse the XML, they might be treating either "appendix" or "app." as the long form. For users treating "app." as the long form, the fix would be changed behaviour.

I think it's worth doing anyway, but wanted to flag the possible consequences. Thanks for maintaining the CSL repos.

### DIFF
--- a/locales-en-GB.xml
+++ b/locales-en-GB.xml
@@ -309,39 +309,39 @@
     </term>
 
     <!-- SHORT LOCATOR FORMS -->
-    <term name="appendix">			 
+    <term name="appendix" form="short">			 
       <single>app.</single>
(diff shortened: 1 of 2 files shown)


adam3smith:
Thanks. These are all new terms as of 1.0.2. I think it's unlikely they have seen much use (not available in Zotero yet, say)

PR #256, opened 2022-03-24
update for 1.0.2

fflamerie:

### Description

- non-breaking spaces before colon and semicolon added
- new terms translated
- some existing translations changed for more consistency with new terms
- not sure about the translation of the locator "elocation"

### Checklist

### DIFF
--- a/locales-fr-FR.xml
+++ b/locales-fr-FR.xml
@@ -4,6 +4,9 @@
     <translator>
       <name>Grégoire Colly</name>
     </translator>
+    <translator>
+      <name>Collectif Zotero francophone</name>
+    </translator>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
     <updated>2012-07-04T23:31:02+00:00</updated>
   </info>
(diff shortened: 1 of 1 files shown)


bwiernik:
Thanks! Just want to check, is line 72 "in" correct? Not "dans" or "à" ?

fflamerie:
Thanks for your feedback! 

- We have not changed line 72. I do not know what would be the best option. It looks like about 60 styles have set  in a `<locale xml:lang="fr">` `<term name="in">dans</term>`. So I guess the majority (173 styles entitled "français") prefer "in".
- line 160 and 161 : BCE and CE would be more appropriate, I'm sorry I've just checked that.

fflamerie:
I apologize for the delay. 
I have edited the file : line 160 and 161 : translations set back to BCE an CE

bwiernik:
For BCE and CE, shouldn't they be 

1. AEC or AÈC or av. è. c.
2. EC or ÈC or è. c.

Cf. https://fr.wikipedia.org/wiki/%C3%88re_commune

@adam3smith The AD/BC equivalents are most common in French. I see examples of both AEC and av. è. c. in use, with AEC seeming more common. Is it worth recording both long and short forms of these terms in the default locale, or should we just leave changing it to the rare style where that would be necessary?

fflamerie:
I have checked with an ancient history specialist and an historian : 

- ‌av. n. è.
- n. è. 

are more most common, I will update the file consequently. An option without diacritics would be possible, _i. e._ 

- av. n. e.
- n. e.

bwiernik:
Awesome thanks for checking on that

PR #260, opened 2022-06-17
Update locales-nl-NL.xml

denismaier:
changes according to https://github.com/jgm/citeproc/issues/109 Can't tell which one is correct, though... @rmzelle ?

### DIFF
--- a/locales-nl-NL.xml
+++ b/locales-nl-NL.xml
@@ -59,10 +59,10 @@
     <term name="circa" form="short">c.</term>
     <term name="cited">geciteerd</term>
     <term name="edition">
-      <single>druk</single>
-      <multiple>drukken</multiple>
+      <single>editie</single>
+      <multiple>editie</multiple>
     </term>
-    <term name="edition" form="short">dr.</term>
+    <term name="edition" form="short">ed.</term>
     <term name="et-al">e.a.</term>
     <term name="forthcoming">in voorbereiding</term>
     <term name="from">van</term>


bwiernik:
Both seem to be in common use? Eg, this page uses both seemingly interchangeably? https://www.scribbr.nl/apa-stijl/wijzigingen-7e-druk/

denismaier:
WDYT @ffel?

denismaier:
Maybe that change could be made in some specific styles only?

adam3smith:
I'm guessing the difference might be similar to Auflage (druk) and Ausgabe (editii) in German? 
In German we ended up going with Auflage because it's more commonly used in citations. 

We actually have "druk" specifically hardcoded in several styles such [as deBuck](https://www.zotero.org/styles/de-buck?source=1) and the [Utrecht history guidelines](https://www.zotero.org/styles/universiteit-utrecht-onderzoeksgids-geschiedenis?source=1)
and it seems expected in others, so I'd be inclined to leave it as-is, unless there's strong evidence to the contrary (e.g., multiple Dutch APA guides suggesting editie)

bwiernik:
A Dutch colleague says both are used but editie is now more common

ffel:
Hi

I teach and I am not too bothered students buy another "druk" (imprint; a
book coming from a next batch of 1000 copies); but I don't like the case
students buy the first "editie", while I use the second "editie" in class
(edition, an updated and possibly extended version of the contents). So an
english book which says "Tour of C++, 2nd edition" on the cover page is
really another book (updated contents) compared to the first edition. I'd
like to reflect the edtion part in the (otherwise Dutch) bibliograph (over
a reference to "druk").

cheers

Chris

Op vr 17 jun. 2022 om 16:11 schreef Denis Maier ***@***.***>:

bwiernik:
I looked at a couple of Dutch adaptations of APA guides ([1](https://www.amsterdamuas.com/library/shared-content/news/2021/07/new-dutch-apa-guidelines-now-available.html), [2](https://www.scribbr.nl/handleiding-apa-regels/)), and they say to use whichever is shown on the book itself. Druk seems to be used in more of their examples, but both are very common. Phrases like "Herziene druk" for "revised edition" are common, and "druk" also seems common to refer numeric editions (e.g., https://www.barnesandnoble.com/w/scrum-wegwijzer-2de-druk-van-haren-publishing/1141391028). So, it doesn't seem like "druk" has a clear meaning as "printing" versus an edition with different content. 

Given that, I'd suggest leaving "druk" in place and letting specific styles specify "editie" or rely on users to enter the edition text with the correct term in their data (e.g., that would be how to fit the guidelines [here](https://www.amsterdamuas.com/library/shared-content/news/2021/07/new-dutch-apa-guidelines-now-available.html)).

ffel:
Thanks for looking into it.

The "APA richtlijnen toegelicht" has "3e herziene editie" on its front
cover (not "druk"). The afore mentioned "A Tour of C++" has "second
edition" on the cover and "first print" in its page left of the table of
contents (don't know its English, my Dutch is better).

I don't like the fact that pandoc citeproc translates "edition: 2" into "2e
druk" (as this obviously wrong for this case as it is the first printing).
I've looked into the CSL 1.0.2 spec: it knows numeric field "edition", I
failed to find a field for "printing" (or, "imprint"). I still think that
the English "edition" is more like the Dutch "editie" than the Dutch "druk".

Nevertheless, I apparently have to accept that many use the field "edition"
to enter the "print"...

It's a pity, still, again, thanks for your attention.


Op vr 17 jun. 2022 om 19:58 schreef Brenton M. Wiernik <
***@***.***>:

ffel:
The APA explanation refered to earlier states that in Dutch bibliography it
is allowed to add "druk" (print) or "editie" (edition). These are, however,
two different values. These are not two Dutch words for the same concept.
For that reason, I think that it is no good idea to translate the English
"edition" by the Dutch "druk".. :)

cheers again

Op vr 17 jun. 2022 om 20:32 schreef chris van uffelen ***@***.***

bwiernik:
I see the difference in meaning (edition vs printing), but it seems like druk is nevertheless often used to mean "edition" rather than printing more narrowly?

ffel:
there is a clear distinction. For example the iconic Dutch dictionary
"dikke van Dale", (thick van Dale,
https://webwinkel.vandale.nl/van-dale-groot-woordenboek-van-de-nederlandse-taal-9789460775963)
is in its "jubileum edition" (honary edition, 16th according to
https://nl.m.wikipedia.org/wiki/Van_Dale_Groot_woordenboek_van_de_Nederlandse_taal)
but first print according to webshop link.

however, it may very well be that Dutch book version were traditionally
often distinguished on "druk" (or "herziene druk", revised print) and that
using edition for "revised print" is more like a anglon tradition"

so short: druk means print, period. However when you add one to it on every
new imprint it is a sufficient good number to prevent students from buying
a too old version (buy at least druk 4). Nowadays even Van Dale uses
edition to tempt us Dutch to buy another kg of printed paper.

one edition can have several printings, even Van Dale hopes they will sell
that huge numbers...

adam3smith:
> The result is that I cannot use "edition" in my Dutch text (I teach and
write a lot of student material in Dutch) with the correct reference to
English works (say, "K&R The C Programming Language (second edition)",
which prints as "second imprint" in Dutch which definitely is something
else.

As I said, you can adjust terms individually in citation styles. As evidenced by the reference to the 7e druk of the APA manual referenced above, Dutch people do not, in fact, all agree with you that 7th edition can't be translated to 7th druk.

Issue #268, opened 2022-09-21
Duplicate terms: "book", "chapter", etc


zepinglee:
I just find `book` and `chapter` terms are used as both item types and locators in locale files. For example:

https://github.com/citation-style-language/locales/blob/62dcca65d754b5e603750838e909a5e90c32e3cc/locales-en-US.xml#L108

https://github.com/citation-style-language/locales/blob/62dcca65d754b5e603750838e909a5e90c32e3cc/locales-en-US.xml#L248-L251

The duplicate term definition may cause unexpected result in citeproc output. Here is the full list of duplicate terms: `interview`, `book`, `chapter`, `book,form:short`, and `chapter,form:short`.

Additionally, PR #254 fixed missing `form="short"` attribute in only `en-US` and `en-GB` but the missing attributes in other languages cause duplicates. I'm happy to make a PR to resolve them.

bwiernik:
Ah, yeah, that's right. They should be removed from the item types list and a note like the one for figure added in their places

bwiernik:
Not sure what you mean. If they are defined in the other locales, they should be removed there too

denismaier:
Sure. But can we automate this process?

denismaier:
Yes. I don't remember, but was there a specific reason we did not do that for `figure`?

zepinglee:
> The item type and locator "figure" are literally the same object. The same would apply to chapter presumably. Book, I'm not sure--in German, if I cited "author, year, Book 2" is that the same term as for "a book"?

It seems that the "book" locator is only used in legal areas like `article` in  (e.g., [Civil Code of China](https://sclx.pkulaw.com/en_law/aa00daaeb5a4fe4ebdfb.html)).

denismaier:
> The item type and locator "figure" are literally the same object. The same would apply to chapter presumably. Book, I'm not sure--in German, if I cited "author, year, Book 2" is that the same term as for "a book"?

You mean as "a book" as opposed to e.g. "a webpage"? Then yes, it would be "Buch" in both cases.

denismaier:
But that should be solved by #269 right?

PR #269, opened 2022-09-22
Resolve duplicate terms

zepinglee:
The duplicate term issue was discussed in #268 and this is a proposed PR.

I've also written a Python3 script to search for duplicate terms in each locale file and found several ones. I then resolved them in 18f3eb5 (thanks to Google Translate).

```python
import glob
import xml.etree.ElementTree as ET

ns = {'cs': 'http://purl.org/net/xbiblio/csl'}


def check_duplicates(path):
    tree = ET.parse(path)

    terms_dict = dict()
    duplicates = []

    for term in tree.findall('.//cs:term', ns):
        term_key = ' '.join([
            f'{key}="{value}"'
            for key, value in sorted(term.attrib.items(),
                                     key=lambda x: "0"
                                     if x[0] == 'name' else x[0])
        ])
        if term_key in terms_dict:
            duplicates.append(term_key)
        terms_dict[term_key] = term.text

    if duplicates:
        print(path)
        for term_key in duplicates:
            print('    ' + term_key)


for path in sorted(glob.glob('locales-*.xml')):
    check_duplicates(path)
```

### DIFF
--- a/locales-af-ZA.xml
+++ b/locales-af-ZA.xml
@@ -66,7 +66,6 @@
     <term name="in">in</term>
     <term name="in press">in press</term>
     <term name="internet">internet</term>
-    <term name="interview">interview</term>
     <term name="letter">brief</term>
     <term name="no date">no date</term>
     <term name="no date" form="short">n.d.</term>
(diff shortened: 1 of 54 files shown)


bwiernik:
Thanks!

PR #274, opened 2023-01-12
Fix date format for Basque

mirenbz:
### Description

Dates are not rendering in correct Basque, which requires different formats for the `date-bib` and `access` macros (retrieved date). I've tried to modify date elements so they can both be generated correctly. This will require some modifications in the `.csl` file as well, which I'm working on: https://github.com/citation-style-language/styles/pull/6370.

I have also corrected some typos. 

### Checklist

### DIFF
--- a/locales-eu.xml
+++ b/locales-eu.xml
@@ -12,7 +12,7 @@
   <date form="text">
     <date-part name="year" suffix="(e)ko "/>
     <date-part name="month" suffix="aren "/>
-    <date-part name="day" suffix="a"/>
+    <date-part name="day" suffix="(e)an"/>
   </date>
   <date form="numeric">
     <date-part name="year" suffix="/"/>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! Out of curiosity, what's the role of the parentheses here? Do they indicate alternate forms (like e.g. profesor(a) in Spanish) or are they regular punctuation?

mirenbz:
Hi @adam3smith , sorry I missed your comment! The parentheses are to indicate alternate forms to attach the suffix, depending on the last letter of the word. If the word ends with a consonant, then the suffix takes the form "ean", but if it ends with a vowel, it's "an". So "(e)an" is used when the preceding word is variable and can end in either a vowel or a consonant. 

As in this case it's used for the day of the month, for example, 1 in Basque is "bat", which would transform to "bat**ean**" (_on the 1st_), but 2 is "bi", which should read "bi**an**" (_on the 2nd_).

Hope it helped clarify!

PR #275, opened 2023-01-21
Update nn and nb (Norwegian)

dittaeva:
### Description

Update locales-nn-NO.xml and locales-nb-NO.xml with new terms etc. from new version of CSL.

#### Abbreviations

In the current submitted patch (and possibly in the former version) I think at least half of the abbreviations are invented for CSL and not established forms in use. What is the best way to handle this?

1. Invent abbreviations (as done in patch)?
2. Only include established abbreviations, and leave english terms?
3. Only include established abbreviations, and remove english terms?
4. Alternative 2 or 3, but with latin abbreviations where available?
5. Add long form terms instead of abbreviated terms?

#### Multiples

I suspect that the multiple forms of many terms are superflous, or wrong, but can`t be sure, as I`m not sure what it looks like in practice. What could be example uses of multiples of for example "canon", "elocation" or "equation" be?

### Checklist

### DIFF
--- a/locales-nb-NO.xml
+++ b/locales-nb-NO.xml
@@ -203,7 +203,7 @@
     </term>
     <term name="scene">			 
       <single>scene</single>
-      <multiple>scener</multiple>						 
+      <multiple>scene</multiple>						 
     </term>
     <term name="table">			 
       <single>tabell</single>
@@ -419,7 +419,7 @@
     <!-- SHORT ROLE FORMS -->
     <term name="executive-producer" form="short">
       <single>utøv. prod.</single>
-      <multiple>utøv. prodr.</multiple>
+      <multiple>utøv. prod.</multiple>
     </term>
     <term name="producer" form="short">
       <single>prod.</single>
--- a/locales-nb-NO.xml
+++ b/locales-nb-NO.xml
@@ -19,33 +19,33 @@
     <date-part name="day" form="numeric-leading-zeros" prefix="-" range-delimiter="/"/>
   </date>
   <terms>
-    <term name="advance-online-publication">advance online publication</term>
+    <term name="advance-online-publication">forhåndspublisert på nett</term>
     <term name="album">album</term>
-    <term name="audio-recording">audio recording</term>
+    <term name="audio-recording">lydopptak</term>
     <term name="film">film</term>
-    <term name="henceforth">henceforth</term>
+    <term name="henceforth">heretter</term>
     <term name="loc-cit">loc. cit.</term> <!-- like ibid., the abbreviated form is the regular form  -->
-    <term name="no-place">no place</term>
-    <term name="no-place" form="short">n.p.</term>
-    <term name="no-publisher">no publisher</term> <!-- sine nomine -->
-    <term name="no-publisher" form="short">n.p.</term>
-    <term name="on">on</term>
+    <term name="no-place">intet sted</term>
+    <term name="no-place" form="short">i.s.</term>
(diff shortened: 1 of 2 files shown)


bwiernik:
Thanks! If a short form isn't used in practice, please remove it so that the long form will be used instead by default. 

For the plural forms, for example, if I were to cite equations 4–6, or locations 15–92, or cannons 2–3.

dittaeva:
Ok, please hold off on the pull request then, I'll do some changes first.

dittaeva:
If singular and plurals/multiples are the same, shall I still include both using the same form/word, or can I remove the singular/multiple tags?

I understand that the terms will be used the same was as "pages 3–10" for example, not much more complicated than that.

bwiernik:
Please specify singular and plural terms even if they are the same

dittaeva:
So, for the short forms that aren't used in practice, do I remove the whole term (f.ex. " <term name="no-place" form="short">i.s.</term>"), just the translation ("i.s."), or reinsert the english template ("n.p.")?

dittaeva:
BTW: the short forms of both no-place and no-publisher is "n.p.", and I hesitate making a report for it, them being so close together. Are they perhaps separated out of the same previous term? The way it is now, one could end up with "n.p. n.p." in the citation I suppose. Doesn't seem ideal.

bwiernik:
Under no circumstances leave the English term unless that is actually used in the locale. 

If there is no commonly used abbreviation, delete the whole short form term. 

It's fine if no-publisher and no-place have the same abbreviation. That's common in English styles unless the Latin sine nomine and sine locale are used

dittaeva:
What about the multiples of long role forms, such as "chair", are they also used "chairs 4–6" f.ex?

adam3smith:
Roles would be used in citations, e.g. of a conference panel.

Wiernik, B. & Karcher, S. (chairs), Best practices in formatting citations [panel], _Annual meeting of the Citation Style Society,_ Cancún, Mexico, April 1-3, 2022.

dittaeva:
Thanks. The pull request should be ready now. Please excuse all the commits, I can redo it if necessary, but am somewhat limited as I'm working directly through the github web interface.

adam3smith:
(we can squash commits, no worries)

bwiernik:
Thanks and great work!

dittaeva:
> @dittaeva --[ we've gotten a complaint about using "ss." instead of "s."](https://forums.zotero.org/discussion/103826/question-about-double-ss-pp-for-multiple-pages-in-norwegian-apa7#latest) for page ranges. The post is specific to APA Norsk style, but given how widely used that is, are you sure that "ss." is really the better (in the sense of: more widely used) abbreviation?

I honestly don't know why I introduced "ss.". However, as is pointed out the [Norwegian APA manual (which seems to be unofficial)](https://old.unit.no/sites/default/files/media/filer/2019/01/APA-norsk.pdf) does only use "s.", and more importantly [the Language Council of Norway explicitly prescribes "s." also for multiple pages](https://www.sprakradet.no/sprakhjelp/Skriveregler/Forkortinger/), although that is for all usage cases, not citation in particular.

I'd be grateful if you could take care of it @adam3smith

adam3smith:
will do -- thank you!

dittaeva:
For both nn and nb

adam3smith:
Fixed in https://github.com/citation-style-language/locales/commit/2919a84bffbae974a2702be0f5fa2ff7a6a51859

Issue #169, opened 2017-12-11
Make composer package - Feature request


Deyter:
**Motivation**
We using your project in our PHP project.
The best way to manage PHP(and any other) project requirements is package checking/updating.
The most popular package manager is Composer.
**Potential solving**
Can you make composer package from your project?
Example of composer.json file which must be in the root folder of project attached(archived for attaching).
[composer.json.zip](https://github.com/citation-style-language/locales/files/1546942/composer.json.zip)

rmzelle:
We're not opposed to accepting this, but we first need to figure out how to add a "composer.json" file to the "styles-distribution" repo.

There is discussion for that at https://github.com/citation-style-language/styles/pull/2706 and an incomplete pull request at https://github.com/citation-style-language/distribution-updater/pull/3. If you could help us fix up that PR, we'd consider also adding a "composer.json" file to the "locales" repo.

rmzelle:
As background, https://github.com/citation-style-language/styles-distribution hosts a cleaned-up copy of https://github.com/citation-style-language/styles, and is only updated by https://github.com/citation-style-language/distribution-updater if the Travis CI tests on the "styles" repo complete without errors.

I therefore think it makes most sense to add a "composer.json" file to the "styles-distribution" repo (since downstream users should usually pull their styles from there), but this would require some changes to "distribution-updater". The updater isn't my code, and I'm not the best coder, so any help with preparing a good PR would be much appreciated. Once we have that situation figured out adding a composer.json to the "locales" repo would be easy.

bkosborne:
This would be helpful for us. Looks like https://github.com/citation-style-language/styles is already set up this way, and a git tag is created for each commit which is helpful to, so it allows people to "lock" to a specific version of the repo.

AS-adci:
Hi @adam3smith.
I can see that GitHub workflow already copied here from styles repository.
So you just need to add to repository composer.json file.
You can use [file from styles](https://github.com/citation-style-language/styles/blob/master/composer.json) as template or I can create PR later.

Deyter:
It's good enough for this purpose
https://github.com/citation-style-language/locales/pull/161

Deyter:
@adam3smith
Could you also create the package for project like you made it for styles?
https://packagist.org/packages/citation-style-language/styles
Then this issue can be marked as finished.

adam3smith:
[...]

PR #281, opened 2023-03-23
correct license to match the SPDX license identifier.

seboettg:
### Description

use SPDX conform term for license. See issue #169


adam3smith:
Thanks!

PR #286, opened 2023-06-27
Create locales-gl-GL.xml

mmgn897:
New translation to Galician language


bwiernik:
Thanks so much for the contribution! As far I can see looking at ISO locales lists, the name of the locale is typically written as "gl-ES". Does that sound correct to you @mmgn897 ?

bwiernik:
Oh, I see you confirmed that gl-ES was fine in your other PR https://github.com/citation-style-language/locales/pull/225

PR #288, opened 2023-08-09
Sr latn rs locale translation

nikola-grubor:
### Description

As Serbian has two official scripts, the Latin script being used for scientific writing, I propose changing "sr-RS" (which was in Cyrillic) to "sr-Cyrl-RS" as per BCP 47.

I have also partially translated the existing "sr-RS" locale file into its Latin script equivalent ("sr-Latn-RS").

### Checklist


bwiernik:
@adam3smith How do you think we should handle this? I think it should work fine to have to have a locale with the script indicated ala "sr-Cyrl-RS" as far as processors are concerned, so we just need to modify our test regex. Does that sound good to you?

PR #289, opened 2023-08-12
Update locales-pt-BR.xml

Eskeminha:
### Description

Updated a lot of Locales. FYI, there are a lot of portguese terms that have no short forms, if you want me to, I can change the ones for their normal use.

### DIFF
--- a/locales-pt-BR.xml
+++ b/locales-pt-BR.xml
@@ -216,8 +216,8 @@
       <multiple>artigos</multiple>						 
     </term>
     <term name="canon">			 
-      <single>canon</single>
-      <multiple>canons</multiple>						 
+      <single>cânon</single>
+      <multiple>cânones</multiple>						 
     </term>
     <term name="elocation">			 
       <single>localização</single>
--- a/locales-pt-BR.xml
+++ b/locales-pt-BR.xml
@@ -7,8 +7,11 @@
     <translator>
       <name>Meira da Rocha</name>
     </translator>
-        <translator>
+    <translator>
       <name>Renato Cirino</name>
+    </translator>
+    <translator>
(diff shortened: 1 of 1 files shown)


bwiernik:
@adam3smith for terms that don't have a short form in a langauage, I'm thinking it would be easier to maintain to duplicate the long term in the short form. That way we know it's intentional and not missing. What do you think?

adam3smith:
> @adam3smith for terms that don't have a short form in a langauage, I'm thinking it would be easier to maintain to duplicate the long term in the short form. That way we know it's intentional and not missing. What do you think?

Agreed. Will also probably make scripting any bulk updates easier

bwiernik:
@Eskeminha for any terms that don't have a short form, could you add short forms with the same text as the long form?

Thanks so much for your contribution!

Eskeminha:
@bwiernik done.

I just have a final question about one thing, this one:

`218 	<term name="canon">			 
      		<single>canon</single>
		<multiple>canons</multiple>						 
	</term>`
	
Are these related to the biblical canons, or what? I am trying to figure out so I can check for the correct translation.

bwiernik:
Yes, biblical canons and similar for other classic works

Eskeminha:
> Yes, biblical canons and similar for other classic works

Ok, lemme do this final change then.

bwiernik:
@Eskeminha did you intend to close this?

Eskeminha:
Sorry, was cleaning some repos and prob this one went with it :(

Eskeminha:
@bwiernik Do you want me to recreate it, or can you still merge this?

bwiernik:
Can you recreate it?

Eskeminha:
Sure I can

PR #291, opened 2023-09-20
Add verb form of term "hearing"

bwiernik:
@adam3smith What do you think of this?

"hearing" is an item type term rather than a role term, but it makes sense to me to use the verb form of the term to refer to the author/giver of testimony at the hearing

This would be used for localizing phrasing like `(testimony of <author>)`


bwiernik:
@adam3smith I'm done adding things here for now.

bwiernik:
@adam3smith Are you okay with merging this?

adam3smith:
The only controversial bit is hearing I guess, but why not?

PR #292, opened 2023-09-28
Update Japanese Localization

cmplstofB:
### Description

Update Japanese localization file (locales-ja-JP.xml)

- Follow CSL 1.0.2 (issue #246)
- Remove unnecessary plural form definitions. (In Japanese, it is almost always singular and plural isomorphic)
- Solve issue #237 (line 304)

### Checklist


POBrien333:
@cmplstofB 
I've just resolved the merge conflicts as in the meantime new terms were added to the locales.
Can you go over the file and add any missing translations as well as checking if I have introduced any errors?
We appreciate your help!

PR #301, opened 2024-03-12
Translate v1.0.2 terms for zh-CN

zepinglee:
### Description

The Chinese (zh-CN) translations of v1.0.2 terms for are added. The terms that already have Chinese translations are not touched.

### Checklist


zepinglee:
I find the terms in en-US locale were reordered in 9d5a790 so that I can't compare zh-CN locale with en-US one. How about writing a script to apply the same sorting method to other locales?

PR #303, opened 2024-03-17
Wrong translation for "Online" pt-PT

chicoferreira:
### Description

The translation found in https://github.com/citation-style-language/locales/blob/master/locales-pt-PT.xml#L91 for "Online" is "Em linha" which is wrong. "Em linha" is the literal translation for "On line", but there is no translation for many foreign words, including "Online".


bwiernik:
Thanks!

PR #305, opened 2024-03-21
Add locales-bal-PK.xml (Balochi – Pakistan)

bgo-eiu:
### Description

This pull request adds a locale for Balochi (Pakistan). Some additional change may be necessary to accommodate the three letter language code.

### Checklist

### DIFF
--- a/locales-bal-PK.xml
+++ b/locales-bal-PK.xml
@@ -1,0 +1,478 @@
+<?xml version="1.0" encoding="utf-8"?>
+<locale xmlns="http://purl.org/net/xbiblio/csl" version="1.0" xml:lang="pa-PK">
+  <info>
+    <translator>
+      <name>عثمان</name>
+    </translator>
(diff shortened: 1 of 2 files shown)


locales-bal-PK.xml
```diff
@@ -0,0 +1,478 @@
+<?xml version="1.0" encoding="utf-8"?>
+<locale xmlns="http://purl.org/net/xbiblio/csl" version="1.0" xml:lang="pa-PK">
```
bwiernik, line 2: The locale name needs to be updated here

bwiernik (changes requested):

bwiernik:
@adam3smith I think we need to update our filename checking regex here. Balochi does not have a 2-letter ISO code, only a 3-letter code, so the filename check is currently failing.

adam3smith:
Changing the Regex is simple of course, but I'd like to make sure three-letter language codes won't confuse processors

bwiernik:
Bumping this to top of list since we have a second 3 letter locale PR

PR #306, opened 2024-03-31
Create locales-brh-PK.xml (Brahui – Pakistan)

bgo-eiu:
### Description

Brahui is a language used in Pakistan; the terminology has some overlap with that of other Pakistani languages. (I intend to assemble some bibliographies of linguistics resources in Pakistani regional languages, so there are more of these translations to come.)

### Checklist


bwiernik:
Thanks for doing all of these! For the languages that have 3-letter codes, we need to run some tests to be sure that various parts of CSL infrastructure will work with them and not need updates. Up until now all of the locale files have had 2-letter codes. So it might take a bit of time for us to merge these 3-letter locales. We really appreciate your efforts to do the translations and expand CSL support for this region!

PR #309, opened 2024-05-05
Update locales-fr-FR.xml

POBrien333:
Just came across this working on a custom style. I do not think "in" is French at all.

### Description

Briefly describe the changes you're proposing.

### Checklist

### DIFF
--- a/locales-fr-FR.xml
+++ b/locales-fr-FR.xml
@@ -85,7 +85,7 @@
     <term name="forthcoming">à paraître</term>
     <term name="from">à l'adresse</term>
     <term name="ibid">ibid.</term>
-    <term name="in">in</term>
+    <term name="in">dans</term>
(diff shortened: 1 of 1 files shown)


adam3smith:
I think that is on purpose -- my recollection is that French academics more commonly use the Latin "in" (which happens to be the same as the English one). "dans" is in the French Canadian locale: https://github.com/citation-style-language/locales/blob/master/locales-fr-CA.xml#L85

bwiernik:
Do we have a French contributor we can verify with?

POBrien333:
Maybe @fflamerie or @gracile-fr would know?

bwiernik:
I checked four continental French and Belgian psychology journals and they all used "In"

POBrien333:
Ok, thank you so much. I guess we can close this as a false alarm. :)

PR #314, opened 2024-06-27
Patch 1

natalia-legarska:
### Description

The changes I'm proposing are based on APA7 standards for Polish language:

Skimina, E., Harasimczuk, J., Cieciuch, J. (2022). Podstawowe standardy edytorskie naukowych tekstów psychologicznych w języku polskim na podstawie reguł APA 7. Wydawnictwo Liberi Libri. https://doi.org/10.47943/lib.9788363487560 

according to that resouce for websites the correct format is instead of: 
`Doe, J. (b.d.). Article title. Pobrano 27 czerwiec 2022, z https://mywebsite.org/my-article`
it should be:

`Doe, J. (b.d.). Article title. Pobrane 27 czerwca 2022, z: https://mywebsite.org/my-article`

therefore I needed to fix long month forms, `from` and `retrieved` fields.

### Checklist

### DIFF
--- a/locales-pl-PL.xml
+++ b/locales-pl-PL.xml
@@ -11,7 +11,7 @@
       <name>Pendzoncymisio</name>
     </translator>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
-    <updated>2012-07-04T23:31:02+00:00</updated>
+    <updated>2024-06-27T15:28:46+02:00</updated>
   </info>
   <style-options punctuation-in-quote="false"/>
   <date form="text">
@@ -106,7 +106,7 @@
     </term>
     <term name="review-of">review of</term>
     <term name="review-of" form="short">rev. of</term>
-    <term name="retrieved">pobrano</term>
+    <term name="retrieved">pobrane</term>
     <term name="scale">skala</term>
     <term name="version">wersja</term>
 
@@ -657,19 +657,19 @@
     <term name="editortranslator" form="verb-short">red.tłum.</term>
 
     <!-- LONG MONTH FORMS -->
-    <term name="month-01">styczeń</term>
-    <term name="month-02">luty</term>
-    <term name="month-03">marzec</term>
-    <term name="month-04">kwiecień</term>
-    <term name="month-05">maj</term>
-    <term name="month-06">czerwiec</term>
-    <term name="month-07">lipiec</term>
-    <term name="month-08">sierpień</term>
-    <term name="month-09">wrzesień</term>
-    <term name="month-10">październik</term>
-    <term name="month-11">listopad</term>
-    <term name="month-12">grudzień</term>
-
+    <term name="month-01">stycznia</term>
+    <term name="month-02">lutego</term>
+    <term name="month-03">marca</term>
+    <term name="month-04">kwietnia</term>
+    <term name="month-05">maja</term>
+    <term name="month-06">czerwca</term>
+    <term name="month-07">lipca</term>
+    <term name="month-08">sierpnia</term>
+    <term name="month-09">września</term>
+    <term name="month-10">października</term>
+    <term name="month-11">listopada</term>
+    <term name="month-12">grudnia</term>
+    
     <!-- SHORT MONTH FORMS -->
     <term name="month-01" form="short">sty.</term>
     <term name="month-02" form="short">luty</term>
--- a/locales-pl-PL.xml
+++ b/locales-pl-PL.xml
@@ -10,6 +10,9 @@
     <translator>
       <name>Pendzoncymisio</name>
     </translator>
+    <translator>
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks -- note that these changes would apply to all styles, so this is probably not the right place (and definitely isn't for the z:) 
Could you explain the difference between the months long forms and pobrano/pobrane? 
For the z, I would question the Polish translation you're linking to: the English version of APA doesn't have a colon after ", from" so it seems arbitrary to me to add this in a translation.

natalia-legarska:
@adam3smith Thanks for the reply! 

For months I'm pretty sure it's correct way to decline them, and there are a couple of polish styles that define them in such way. Examples:
- https://github.com/citation-style-language/styles/blob/b2be5aeeee7f00fd2032ac1daad995bbe95398cf/polish-legal.csl#L20
- https://github.com/citation-style-language/styles/blob/b2be5aeeee7f00fd2032ac1daad995bbe95398cf/uniwersytet-kardynala-stefana-wyszynskiego-w-warszawie-przypis.csl#L24
- https://github.com/citation-style-language/styles/blob/b2be5aeeee7f00fd2032ac1daad995bbe95398cf/uniwersytet-kardynala-stefana-wyszynskiego-w-warszawie-autor-rok.csl#L24

I'm aware it's not many examples, but I think there are many people who don't know how to fix the months so they'd go and fix it manually.

As for the z: and pobrane changes, I agree that might be something we don't want to change here, should I then update the APA 7 style with language specific changes? The document I mentioned is an official guide adapted to polish language.

Thanks!

bwiernik:
@kjklosowski Could you potentially weight in?

kjklosowski:
Oh hi thanks for including me to the discussion. My 2 cents:
#### Change to months names
This insight is very good and I have actually observed that issue with couple of Polish styles. See for example [Studia BAS (Polski)](https://www.zotero.org/styles/studia-bas?source=1). The author specifically included citation for webpages, posts etc. (with accessed, date in dd-mm-yyyy formula) but did not fix the issue with the months. Therefore, the issue reported by @natalia-legarska is real and that style has this problem (well the style has this problem with url accessed formula, but that's very smiliar case):

> `Pobrano 27 czerwiec 2022` 

That's a serious grammatical mistake, elementary school-level, see: https://www.ekorekta24.pl/jak-odczytywac-daty/

On the other hand, there is a style [UKSW](https://www.zotero.org/styles/uniwersytet-kardynala-stefana-wyszynskiego-w-warszawie-przypis?source=1) where the author noticed that issue and resolved it in style's local locale:

```
<term name="month-01">stycznia</term>
<term name="month-02">lutego</term>
<term name="month-03">marca</term>
<term name="month-04">kwietnia</term>
<term name="month-05">maja</term>
<term name="month-06">czerwca</term>
<term name="month-07">lipca</term>
<term name="month-08">sierpnia</term>
<term name="month-09">września</term>
<term name="month-10">października</term>
<term name="month-11">listopada</term>
<term name="month-12">grudnia</term>
```
Which is exactly what @natalia-legarska is proposing!

Now, whether we adopt it depends on the logic we follow:
- We can consider that the denominator should be default (because it really is) and, if the author knows their style needs months in genitive case, they should change that locally (like Studia BAS did not and UKWS did),
- Or we can assume that, because in most cases known to me, the dates are printed in either dd-mm-yyyy, we should consider using the genitive case as the most beneficial for most of the people. The only case where it would backfire would be if some style used (uncommon) date ordinal of mm-yyyy. I have never seen something like this, personally. We could add a comment where, if the author wants mm-yyyy, they should change that back in local locale...

Ultimately, that's a serious problem with languages that declinate heavily. No matter what we choose, there will be some edge cases. I would personally go with the change, as imo that's the behaviour most people would expect. As a matter of fact, the declination is a reason I believe some of the terms in the locale are not possible to be translated (the roles terms mostly... but that's another topic).

#### As for the second - pobrane or pobrano
This is a little bit more tricky. Both translate to English to 'downloaded from'. The difference is very nuanced:
- Pobrane z - the object has been downloaded from...
- Pobrano z - the subject has downloaded the object from...

(subject is a person, the object is a thing that has been downloaded ofc)

Despite both of them being very neutral and almost impersonal, I would personally choose the latter option, I think it is used much more widely. At least that's definietly the case in humanities - my branch.

Please let me know what you all think :)

And happy to see a fellow Polish person here @natalia-legarska. CSL and reference managers are so undiscovered and demonised by most of the Polish scholars I deal with...

locales-pl-PL.xml
```diff
@@ -89,7 +92,7 @@
     <term name="edition" form="short">wyd.</term>
     <term name="et-al">et al.</term> <!-- Translated 'i in.' is used less commonly -->
     <term name="forthcoming">w przygotowaniu</term>
-    <term name="from">z</term>
+    <term name="from">z:</term>
```
bwiernik, line 95:
```suggestion
    <term name="from">z</term>
```

PR #319, opened 2024-08-12
Remove verb forms of "author" term

bwiernik:
Styles are generally written expecting these terms to be empty. (cf. https://forums.zotero.org/discussion/116703/zotero-7-inserting-by-into-subsequent-cites#latest)

@adamsmith The author terms were added to zh-TW before they were added to the en-US locale, so I'm not sure whether to keep them. They are not included in the zh-CN locale. What do you think? https://github.com/citation-style-language/locales/commit/3f25b0c9e3c5958fdf054765d05f4f707784f59a

### DIFF
--- a/locales-en-US.xml
+++ b/locales-en-US.xml
@@ -685,7 +685,6 @@
     </term>
 
     <!-- VERB ROLE FORMS -->
-    <term name="author" form="verb">by</term>
     <term name="chair" form="verb">chaired by</term>
     <term name="collection-editor" form="verb">edited by</term>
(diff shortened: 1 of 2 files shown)


zepinglee:
> They are not included in the zh-CN locale.

Yes, the `author` term is not used in most zh-CN (China mainland, simplified Chinese) styles but I know little about `zh-TW` styles.

bwiernik:
I didn't see any examples of this term in a handful of zh-TW articles I checked.

PR #294, opened 2023-10-29
translate locales-de-DE.xml for new CSL 1.0.2 terms

POBrien333:
## CSL Locales Pull Request Template

You're about to create a pull request to the CSL locales repository.
If you haven't done so already, see <http://docs.citationstyles.org/en/stable/translating-locale-files.html> for instructions on how to translate CSL locale files, and <https://github.com/citation-style-language/styles/blob/master/CONTRIBUTING.md> on how to submit your changes.
In addition, please fill out the pull request template below.

### Description

Briefly describe the changes you're proposing.

### Checklist

- [x] Check that you're listed as a `<translator>` in the `<info>` block at the beginning of the file.

PR #329, opened 2025-01-08
Improvements to German-language locales

nchachereau:
### Description

This pull request reverts some changes made to the Swiss German locale in #294 but that were in all likelihood unintended (@POBrien333 could you confirm?). Only the translation of new CSL terms was intended, but some differences between `locales-de-CH.xml` and `locales-de-DE.xml` were erased. The current pull request asks to restore those (typographical) differences: the ß sign is not in use in Switzerland, and other quotation marks are most commonly used.

Note that I did not revert two other apparently unintended changes:
- the translation of the abbreviations of the locators   "line" and "note" (respectively "Z." and "N." instead of "l." and "n." (the new version seems better to me, it looks like the old abbreviations had simply been left untranslated);
- "Bde." for the abbreviated plural of "volume", rather than just "Bd."

@denismaier what do you think?

### Checklist

### DIFF
--- a/locales-de-AT.xml
+++ b/locales-de-AT.xml
@@ -297,8 +297,8 @@
       <multiple>Zeilen</multiple>
     </term>
     <term name="note">
-      <single>Note</single>
-      <multiple>Noten</multiple>
+      <single>Anmerkung</single>
+      <multiple>Anmerkungen</multiple>
     </term>
     <term name="opus">
       <single>Opus</single>
@@ -408,7 +408,7 @@
     <term name="folio" form="short">Fol.</term>
     <term name="issue" form="short">Nr.</term>
     <term name="line" form="short">Z.</term>
-    <term name="note" form="short">N.</term>
+    <term name="note" form="short">Anm.</term>
     <term name="opus" form="short">op.</term>
     <term name="page" form="short">
       <single>S.</single>
--- a/locales-de-AT.xml
+++ b/locales-de-AT.xml
@@ -22,6 +22,9 @@
     <translator>
       <name>Patrick O'Brien</name>
     </translator>
+    <translator>
+      <name>Nicolas Chachereau</name>
+    </translator>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
     <updated>2023-03-28T23:31:02+00:00</updated>
   </info>
--- a/locales-de-CH.xml
+++ b/locales-de-CH.xml
@@ -16,6 +16,9 @@
     <translator>
       <name>Patrick O'Brien</name>
     </translator>
+    <translator>
+      <name>Nicolas Chachereau</name>
+    </translator>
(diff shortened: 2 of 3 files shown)


denismaier:
Yes, ß should be indeed ss. As for the typographical quotes: I don't think institutional guidelines agree on this matter, but at least MS-Word uses «» instead of „“ when the document language is set to "German (Swiss)" so it makes sense to use the same conventions here.

denismaier:
Why do you think "Bde." was unintended?

Z. looks also good to me, but I'm not sure about "N." or about "Note". I think this should actually be "Anmerkung" and "Anm." The German "Note" sounds like describing music, not like the generic form of both "Fussnote" and "Endnote".

nchachereau:
> As for the typographical quotes: I don't think institutional guidelines agree on this matter, but at least MS-Word uses «» instead of „“ when the document language is set to "German (Swiss)" so it makes sense to use the same conventions here.

I don't know about institutional guidelines. The Swiss, German-language, academic books and journal issues I have at home (granted, mostly history books) all use «», as does the NZZ. (The exceptions I could find were works of fiction (e.g. published in Zurich by Diogenes), that use »«.)

> Why do you think "Bde." was unintended?

In the sense that the aforementioned PR #294 only claimed to translate new terms, and this is not a new term. I think @POBrien333 first translated the new terms in `locales-de-DE.xml`, and then probably used a diff to bring the new translations to `locales-de-CH.xml`, thereby also inadvertently erasing some differences between the two locales.

> Z. looks also good to me, but I'm not sure about "N." or about "Note". I think this should actually be "Anmerkung" and "Anm." The German "Note" sounds like describing music, not like the generic form of both "Fussnote" and "Endnote".

Very good point. I agree. That would also apply to the "long" form of the term ("Anmerkung" rather than "Note"), and maybe also to the German-German and Austrian-German locales, and not only the Swiss-German one… what do you want me to do? Only change the `de-CH` file, or all locales? In this PR, or a new one?

bwiernik:
Can you please make all the changes in this PR?

nchachereau:
I had a quick look at `blame` for `locales-de-DE.xml` and it looks like the translation of "note" as "Note" was introduced 16 years ago by @rmzelle, and has not been touched since. It is probably rarely used as a locator, but nevertheless 6 German-language styles* have corrected "N." to "Anm.", so I went ahead and changed the translation in all German locales.

* nccr-mediality.csl, zeithistorische-forschungen.csl, zeitschrift-fur-kunstgeschichte.csl, uni-fribourg-theologie.csl, journal-fur-kunstgeschichte.csl, zeitschrift-fur-theologie-und-philosophie.csl

nchachereau:
(I think the PR should be ready now.)

denismaier:
I think that looks good, but you should add yourself as a translator to all three locales.

nchachereau:
Oh right. I have done so, even if arguably you are the translator here :wink:

adam3smith:
Besten Dank

POBrien333:
Vielen Dank :)

PR #332, opened 2025-02-06
fix(fi-FI): change "vuosikerta" to "volyymi"

lajp:
### Description

Translating volume to "vuosikerta" could be problematic for journals that are published more frequently than once a year.

I suggest using the word "volyymi" instead.

### Checklist


lajp:
> What is the best translation for volumes of a book?

Either "volyymi" or then, more specifically, "nide"/"nidos"

PR #340, opened 2025-04-22
Add Ligurian language translation

jeanm:
Implements support for the Ligurian language.

### Checklist


jeanm:
The tests were failing due to the assumption that language codes always have two letters. I attempted to fix the tests and they pass locally with `rake spec`, but CI isn't picking them up.

bwiernik:
Changes to test spec files need to be made first. I can take care of that

jeanm:
@bwiernik thanks for taking a look. I use [citeproc-py](https://github.com/citeproc-py/citeproc-py) and can successfully load the locale file and produce a bibliography with:

```python
bib_source = CiteProcJSON(yaml_data)
bib_style = CitationStylesStyle("chicago-note-bibliography", locale="lij", validate=False)
bibliography = CitationStylesBibliography(bib_style, bib_source, formatter.html)
citations = [
    # ...
]
for citation in citations:
    bibliography.register(citation)

for item in bibliography.bibliography():
    print(item)
```

Schema validation needs to be disabled (line 2 in the listing above) as `citeproc-py` seems to be using an older schema (other locale files in this repo fail too). However I separately ensured that the new locale file conforms to the schemas in https://github.com/citation-style-language/schema/.

bwiernik:
Awesome thanks (yes it looks like that processor implements 1.0.1, so files with 1.0.2 fields won't validate)

Could you test with pandoc and citeproc-js? These are the most widely used processors and we want to be sure that the 3-letter locales won't cause issues with them

jeanm:
* Pandoc uses [jgm/citeproc](https://github.com/jgm/citeproc) so I downloaded the source, patched the Makefile to pull from my fork of this repo, and built with `cabal configure --flag="executable" && cabal build`. I tested using the `citeproc` executable with `--lang=lij-IT` using the same bibliography I had used with `citeproc-py`, and saw no issues.
* For citeproc-js, I cloned their repo, replaced `locales.json` and `locales-lij-IT.xml` from my branch, and ran the tests with `cslrun -a`, which all passed. I instantiated a new engine with `const citeproc = new CSL.Engine(sys, style, 'lij-IT')`, added all the items, compiled the bibliography with `citeproc.makeBibliography()` and printed the results. No issues.

bwiernik:
Awesome thank you for running those tests!

PR #342, opened 2025-05-29
Update locales-lv-LV.xml

kristoferssolo:
### Description

Added missing translations to fields/values that have been added since 2012.

### Checklist


POBrien333:
Any idea why this is failing? Or is that just a bug with the CI that needs fixing?

POBrien333:
Nice one. Thanks!
I think this one is good to merge as it just translates English to lv-LV with no changes to existing translations.

PR #343, opened 2025-06-23
Update en-US abbreviations and give provenance

adunning:
This update expands CSL's default scholarly abbreviations in `en-US` and ensures that they follow current usage. It does this by checking the provenance of abbreviations in this locale, which is based on the *Chicago Manual of Style*.

Additional abbreviations, especially for short item types, draw first on the [*Oxford Dictionary for Writers and Editors* (*ODWE*)](https://archive.org/details/oxfordstylemanua0000unse), a prescriptive dictionary of which editions before 2005 match US English usage for contractions. Where *ODWE* does not provide guidance, the file draws on the [*Oxford Dictionary of Abbreviations* (*ODA*)](https://doi.org/10.1093/acref/9780199698295.001.0001), a descriptive work.

- Ensure that short forms match the list of scholarly abbreviations in [*The Chicago Manual of Style*, 18th ed. (2024), sec. 10.48](https://www.chicagomanualofstyle.org/book/ed18/part2/ch10/psec048.html):
    - The only prominent change resulting from this is 'trans.' rather than 'tran.' as a singular for `translator`. I cannot any style or dictionary that calls for this: searching for this abbreviation mostly turns up complaints about it from CSL users.
    - Chicago also recommends 'eqq.' as the plural abbreviation for `equation`, which results on one less difference between the `en-US` and `en-GB` localizations.
    - Chicago advocates for 'ca.' rather than 'c.' to abbreviate `circa`, but it seems safer to leave this alone, since they still allow the latter form, and a number of CSL styles use this term; it is unclear how many actively expect the former abbreviation.
    -  Consistently use 'ill.' as the abbreviation for illustrator/illustrated rather than mixing it with 'illus.' The latter is also common, such as in APA and *ODWE*, it would thus also be legitimate to standardize on 'illus.'.
- Remove 'w.' as an abbreviation for 'with', previously used in the `contributor` and `guest` short forms. Styles are likely to call these inadvertently when adding a blanket representation of contributor labels in short forms. This seems too specialized to be a default, since many dictionaries do not give a definition for 'w.' I cannot find a style manual that calls for it.
- Use the recommended month abbreviations from *CMOS* 10.44, which is also _ODWE_ usage.
- Provide missing `editor-translator`. More styles probably need 'ed. and trans.' than 'ed. & trans.', but this is probably not something that CSL can easily change at this point.
- Provide more short forms of general terms and item types, mostly based on _ODA_, and within their own sections of the file for internal consistency.
- Change unclear 'print.' abbreviation to 'ptg.', from the *Oxford Dictionary for Writers and Editors* (no CSL styles yet use it).
- Use terms for `collection` that better describe its use in the CSL specification (from APA).
- Abbreviate `canon` using the less specialized 'can.' rather than 'c.' (*ODWE*).
- Use 'issue' rather than 'number' as long text for `issue`.
- Remove unclear 'cre.' abbreviation for `series-creator`; I cannot find either a source for this or an alternative.
- Provide more descriptive labels for `classic` (classical work) and `collection` (archival collection) to follow the CSL specification.
- Trim whitespace and add contributor IDs.

### DIFF
--- a/locales-en-US.xml
+++ b/locales-en-US.xml
@@ -152,7 +152,7 @@
     <term name="webpage">webpage</term>
 
     <!-- SHORT ITEM TYPE FORMS -->
-    <term name="article-journal" form="short">journal art.</term>
+    <term name="article-journal" form="short">jour. art.</term> <!-- ODWE -->
     <term name="article-magazine" form="short">mag. art.</term> <!-- ODWE -->
     <term name="article-newspaper" form="short">newspaper art.</term>
     <!-- book is in the list of locator terms -->
--- a/locales-en-US.xml
+++ b/locales-en-US.xml
@@ -425,7 +425,7 @@
     <term name="title-locator" form="short">
       <!-- from Oxford Dictionary for Writers and Editors -->
       <single>tit.</single>
-      <multiple>tits.</multiple>
+      <multiple>titt.</multiple> <!-- source unknown for plural -->
     </term>
     <term name="verse" form="short">
       <single>v.</single>
--- a/locales-en-US.xml
+++ b/locales-en-US.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <locale xmlns="http://purl.org/net/xbiblio/csl" version="1.0" xml:lang="en-US">
+  <!-- This file follows the The Chicago Manual of Style, 18th ed. (2024), especially the list of scholarly abbreviations in sec. 10.48; additional abbreviations are from the Oxford Dictionary for Writers and Editors (ODWE) <https://archive.org/details/oxfordstylemanua0000unse>. -->
   <info>
     <translator>
       <name>Andrew Dunning</name>
@@ -44,7 +45,7 @@
     <term name="available at">available at</term>
     <term name="by">by</term>
     <term name="circa">circa</term>
-    <term name="circa" form="short">c.</term>
+    <term name="circa" form="short">ca.</term>
     <term name="cited">cited</term>
     <term name="et-al">et al.</term>
     <term name="film">film</term>
@@ -146,20 +147,22 @@
 
     <!-- SHORT ITEM TYPE FORMS -->
     <term name="article-journal" form="short">journal art.</term>
-    <term name="article-magazine" form="short">mag. art.</term>
+    <term name="article-magazine" form="short">mag. art.</term> <!-- from ODWE -->
     <term name="article-newspaper" form="short">newspaper art.</term>
     <!-- book is in the list of locator terms -->
     <!-- chapter is in the list of locator terms -->
+    <term name="classic" form="short">class.</term> <!-- from ODWE -->
     <term name="document" form="short">doc.</term>
+    <term name="entry-dictionary">dict. entry</term>
+    <term name="entry-encyclopedia">ency. entry</term>
     <!-- figure is in the list of locator terms -->
-    <term name="graphic" form="short">graph.</term>
-    <term name="interview" form="short">interv.</term>
+    <term name="interview" form="short">interv.</term> <!-- source unknown -->
     <term name="manuscript" form="short">MS</term>
-    <term name="motion_picture" form="short">video rec.</term>
-    <term name="report" form="short">rep.</term>
+    <term name="motion_picture" form="short">video rec.</term> <!-- from ODWE -->
+    <term name="report" form="short">rep.</term> <!-- from ODWE -->
     <term name="review" form="short">rev.</term>
     <term name="review-book" form="short">bk. rev.</term>
-    <term name="song" form="short">audio rec.</term>
+    <term name="song" form="short">audio rec.</term> <!-- from ODWE -->
 
     <!-- VERB ITEM TYPE FORMS -->
     <!-- Only where applicable -->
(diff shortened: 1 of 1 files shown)


adunning:
Do we know whether this file tracks particular reference sources? The abbreviation "tran." for a singular translator seems unusual: every style I have checked uses either "trans." or "tr." Could it be an error?

adunning:
> We should probably add this as an implementation detail that processors should implement.

Many thanks indeed for making a note of that!

locales-en-US.xml
```diff
@@ -335,8 +342,9 @@
       <multiple>bks.</multiple>
     </term>
     <term name="canon" form="short">
+      <!-- James A. Brundage, 'The Romano-Canonical Citation System', app. 1 to Medieval Canon Law (Routledge, 1995), https://doi.org/10.4324/9781003156734 -->
       <single>c.</single>
-      <multiple>cc.</multiple>						 
+      <multiple>cc.</multiple>
```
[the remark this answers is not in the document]
bwiernik: For this one, I think let's go with what is more common in classics and religious studies research since that is where these will be cited most often. Is 'c.' or 'can.' more common for those fields?

locales-en-US.xml
```diff
@@ -639,15 +669,20 @@
       <single>ed.</single>
       <multiple>eds.</multiple>
     </term>
+    <term name="editor-translator" form="short">
+      <single>ed. &amp; trans.</single>
+      <multiple>eds. &amp; trans.</multiple>
+    </term>
     <term name="editortranslator" form="short">
-      <single>ed. &amp; tran.</single>
+      <single>ed. &amp; trans.</single>
```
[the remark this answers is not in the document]
adunning: With pleasure! This has been catching me out for years.

locales-en-US.xml
```diff
@@ -248,8 +269,8 @@
       <multiple>folios</multiple>
     </term>
     <term name="issue">
-      <single>number</single>
-      <multiple>numbers</multiple>
+      <single>issue</single>
+      <multiple>issues</multiple>
```
adunning, line 273: @bwiernik Is this change too dangerous? There aren't many styles that use this, but surely one would more normally write out 'volume 1, issue 2' for 'vol. 1, no. 2'. It seems weird that we provide two different ways of getting the text 'number' but none for 'issue'.

bwiernik: Yeah, I think this is reasonable. I rarely (if ever) see "number" spelled out when referring to issue number

adunning:
> For timestamp, could do "at" for long and blank "" for short since most styles use short locator labels?

Good thinking! [Chicago requires text such as 'at 6:09–17'](https://www.chicagomanualofstyle.org/tools_citationguide/citation-guide-1.html#cg-videos) when citing a timestamp. I wonder how many styles really do expect this to be blank; it's not something that many guides even cover. It would be more consistent to provide a label text and expect individual styles to block it, as many do with pages. But I cannot think of an easy way to determine what styles expect.

locales-en-US.xml
```diff
@@ -295,17 +317,15 @@
...
-      <multiple>tables</multiple>						 
-    </term>
-    <term name="timestamp"> <!-- generally blank -->
-      <single></single>
-      <multiple></multiple>						 
+      <multiple>tables</multiple>
     </term>
-    <term name="title-locator">			 
+    <!-- A timestamp is a composite of hours, minutes, etc. and therefore has no default label. -->
+    <term name="timestamp"/>
```
adunning, line 325: I've now looked more closely into the logic of why some styles label a `timestamp` and others do not, since 'generally blank' seemed to have overlooked Chicago's guidance. Fortunately, the logic of the following examples leads to the same conclusion:

- APA 8.13 presents a timestamp following the traditional approach to formatting a canonically numbered part of a work, which is self-labelled by its formatting:
    > (Beck Institute for Cognitive Behavior Therapy, 2012, 1:30:40)
    > (Shakespeare, 1623/1995, 1.3.36–37)

    (The latter citation isn't possible right now, and I've added <https://github.com/citation-style-language/schema/issues/451> to sort this out.)

- Chicago [calls for an 'at' label with a timestamp](https://www.chicagomanualofstyle.org/tools_citationguide/citation-guide-1.html#cg-videos):
    > Cowan, “Green Hydrogen,” at 6:09–17.

    This is necessary because Chicago includes the full length of a film in notes; otherwise, one could end up with '9:15, 6:09'.

- MLA 6.28 also leaves the timestamp unlabelled, though the CSL implementation needs adjustment anyway, because they also require special punctuation:
    > (“Buffy” 00:03:16–17)

   Like APA, this follows their format for canonical references, such as in MLA 6.22.

APA and MLA effectively argue that a timestamp does not need additional labelling because it is has self-describing formatting, following long-established practice for presenting composite reference locators. In this light, I think Chicago's direction to label a timestamp can be understood as a departure that specifically responds to its requirement to include the full duration of a recording, avoiding a potential situation in a note such as '9:15, 6:09–17'. This is not a universal requirement of notes styles. Therefore, I think it is rational to leave `timestamp` unlabelled.

bwiernik:
MLA and APA both call for no label. This guide for film studies students also has no label. https://alexandercollege.ca/web-2024/wp-content/uploads/2024/04/Writing-a-Film-Analysis-2024.pdf

Most biology and medical journals don't do in-text locators, so doesn't matter there.

adunning:
I'm really glad not to have to worry about changing `timestamp`. Thank you for your thoughts!

bwiernik:
Good to merge?

bwiernik:
@adunning is it ready for me to merge today?

adunning:
@bwiernik Yes, I've now sourced all abbreviations and synchronized my changes with the `en-GB` file, so both should be ready to go. Many thanks!

Jason-Abbott:
@adunning , @bwiernik  I am always pleased to see CSL updates!

I noticed this impacts the test suite: https://github.com/citation-style-language/test-suite/pull/75

Is there a standard procedure for updating the suite or a process already underway? I’d be glad to assist ...

PR #351, opened 2025-07-02
Update locales-pl-PL.xml

kjklosowski:
Made couple of changes, mostly to the terms:
- Added translations of no-place, no-publisher terms as they are sometimes used,
- Fixed missing dots,
- Added translations of most locator types. Most of them were left untranslated and some were translated in short form while being left in English in long form. Fixed the consistency and added new translations of least often used ones (like chapter, scene, equation...).

To help the review process, I've marked all my contributions starting with "<!--K.K." and added my reasoning or sources where it was necessary. It should be removed if my changes get accepted!

Enjoy!

PR #353, opened 2025-07-07
Update locales-pl-PL.xml

kjklosowski:
Added translations of:
- first-reference-note-number,
- review-of,

Proposed better translations of:
- personal-communication (komunikacja is more an act of speech, while korespondencja is an object that can be cited).
- no-place (see: "https://www.jezykowedylematy.pl/2023/08/bibliografia-co-zrobic-gdy-brakuje-roku-lub-miejsca-wydania/". This is a fix of my previous commit #351, turns out this last letter is not really used).

Now the most serious one. I would like to propose changing et-al from Polish "i in." back to Latin "et al.". While both Polish and Latin terms are used, from my experience, many publishers require using them consistently. Currently, both ibid and op-cit are in Latin, and only et-al is in Polish. We could either:
- Change all 3 of them to Polish, which would be extreme,
- Change et-al to Latin for consistency, considering Latin the norm, despite locale being Polish (proposed),
- Leave as is, which would be inconsistent.

See: 
- "https://fil.ug.edu.pl/sites/default/files/_nodes/strona-filologiczny/71884/files/kult._prace_standardy_redakcyjne_2.pdf" - page 3,
- "https://wydawnictwo.uwb.edu.pl/wymogi-edytorskie",
- "https://maska.psc.uj.edu.pl/documents/40768330/0/zalecenia+edytorskie.pdf/32ec2a26-8751-451e-8e2a-3d741710a2d3",
- "https://etnologia.amu.edu.pl/wp-content/uploads/2018/04/bibliografia-i-przypisy_zasady.pdf")

### DIFF
--- a/locales-pl-PL.xml
+++ b/locales-pl-PL.xml
@@ -35,13 +35,13 @@
     <term name="henceforth">henceforth</term>
     <term name="loc-cit">loc. cit.</term> <!-- like ibid., the abbreviated form is the regular form  -->
     <term name="no-place">brak miejsca wydania</term>
-    <term name="no-place" form="short">b.m.w.</term>
+    <term name="no-place" form="short">b.m.</term>
     <term name="no-publisher">brak wydawcy</term> <!-- sine nomine -->
     <term name="no-publisher" form="short">b.w.</term>
     <term name="on">on</term>
     <term name="op-cit">op. cit.</term> <!-- like ibid., the abbreviated form is the regular form  -->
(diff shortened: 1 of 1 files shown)


bwiernik:
Based on a quick Google Scholar search for `"i in" source:*.pl` and `"et al" source:*.pl`, it seems that "et al" is about 3x as common as "i in", so let's go with "et al". Can you include the comment `<!-- Translated 'i in.' is used less commonly -->` please?

PR #356, opened 2025-07-28
Update locales-af-ZA.xml

NB-af-ZA:

### Description

Afrikaans has an official list of spelling rules and abbreviations published as "Afrikaans woordelys en spelreëls". This was consulted when translating the af-ZA locale file, which previously contained unrecognised abbreviations and plural forms (eg. “bll” instead of “ble.” for pages). I also made use of all the Afrikaans-language dictionaries, grammar guides and other language resources at my disposal. This locale is done to a professional standard.

The “Guide to Translating CSL Locale Files” states that periods should be included where applicable for abbreviations, and this is done in the locale files I consulted as examples (en-GB and nl-NL), for example short form editor = “ed.”, multiple = “eds.”. In the af-ZA locale file, periods were not included consistently for abbreviations (eg. “ibid.” with period, “ed” without period), which was one of my reasons for translating the locale file properly.

Other examples:
-	Changed the short version of “number” from “no.” to “nr.”. Both are accepted abbreviations, but “nr.” is generally preferred due to the ambiguity of “no.” (which can also refer to “noord-oos”.
-	Changed the short version of “manuscript” from “MS” to the accepted “ms.”. The original “MS” is in Afrikaans the abbreviation for “multiple sclerosis” and not correct in this instance.

I am an editor and translator, and not an IT expert! Therefore, where I believe no relevant Afrikaans term exists, I kept the original line without deleting it and without supplying an Afrikaans term. I chose to do this instead of deleting the line in case someone else wants to attempt translations of these. This is mostly in the section for short verb role forms, which is very rarely used in Afrikaans-language citations.

### Checklist

### DIFF
--- a/locales-af-ZA.xml
+++ b/locales-af-ZA.xml
@@ -16,149 +16,150 @@
     <date-part name="day" form="numeric-leading-zeros" prefix="/"/>
   </date>
   <terms>
-    <term name="advance-online-publication">advance online publication</term>
+    <term name="advance-online-publication">aanlyn voordrukweergawe</term>
     <term name="album">album</term>
-    <term name="audio-recording">audio recording</term>
+    <term name="audio-recording">klankopname</term>
     <term name="film">film</term>
-    <term name="henceforth">henceforth</term>
-    <term name="loc-cit">loc. cit.</term> <!-- like ibid., the abbreviated form is the regular form  -->
-    <term name="no-place">no place</term>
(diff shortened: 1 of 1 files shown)


NB-af-ZA:
Hi @bwiernik, thanks for the quick review! I've added the sources, references where applicable, and fixed two minor issues. Hope I've done it right.

PR #360, opened 2025-08-15
en: Fix ungrammatical 'with guest' as a verb

adunning:
Since 'with guest' is not a verb, rendering the verbal form simply as 'with' ensures grammatical results with an item including more than one `guest`. I cannot find the source for the phrase 'with guest' (it does not occur in any of APA, Chicago, MHRA, MLA, or New Hart's Rules); 'with' is a safer default.


adam3smith:
Good catch, thanks!

PR #366, opened 2025-09-02
Update locales-de-DE.xml (Sept25)

### LINKED ISSUE #162: Short (verb) role editor in German locales
There is a slight discrepancy between the short role and the short verb role term for editors in the German locales. Short role is "Hrsg.", but short verb role is "hg. von". Should this not be harmonized, i.e. either "Hg." and "hg. von" or "Hrsg." and "hrsg. von"? As far as I can tell, "Hrsg."/"Hg." and "hrsg. von"/"hg. von" are equally common.

POBrien333:
Hi @adam3smith and @denismaier and @tillheilmann 
Since I spotted today that some of the German terms were not translated yet, here a suggestion of potential translations. Please make suggestions for any changes and then I can replicate them in the AT/CH locales also.

Also closes #162

(There is also some German citation style PRs that have locale overrides in them since some translations are not available. I'd like to get this PR merged so we can clean up those PRs before merging. E.g. https://github.com/citation-style-language/locales/pull/366)


tillheilmann:
Sorry for not understanding how GitHub really works. Here are my two cents.

I think the following are safe:

```
<term name="article-journal">Zeitschriftenartikel</term>
<term name="article-journal" form="short">Zeitschriftenart.</term>
<term name="entry-encyclopedia">Lexikoneintrag</term>
<term name="report">Bericht</term>
```

"Tonaufnahme" is, I think, more common than "Audioaufnahme", so:

`<term name="song">Tonaufnahme</term>`

'speech' should be:

`<term name="speech">Vortrag</term>`

If you really want to stick with "presentation", it needs the umlaut, like this:

`<term name="speech">Präsentation</term>`

"Standard" is somewhat misleading, the more technical term is "Norm", so maybe better:

`<term name="standard">Norm</term>`

If 'thesis' applies to any kind of degree, then the more fitting would be a generic:

`<term name="thesis">Abschlussarbeit</term>`

"Diplomarbeit" does not include a B.A. or M.A. or doctoral thesis, for example. But then, no one thinks of a doctoral thesis when you say "Abschlussarbeit", so it's difficult.

A conference paper should probably be:

`<term name="paper-conference">Tagungsbeitrag</term>`

But the word "Tagungsbeitrag" also means the paper presentation, not only the published paper, so I am not sure about this. But "Konferenz Paper" is definitely not German and simply wrong.

"Grafik" has a much more narrow meaning than 'graphic', so I would argue for the more general "Bild":

`<term name="graphic">Bild</term>`

Also, "Videoaufnahme" for 'motion_picture' means video and excludes film. The literal translaton of 'motion_picture' is "Film". But then, there already is the term 'film', so...

When it comes to law, German styles are quite quirky, I think, and I am not well versed in these matters. But following appendix III, the right German terms are these:

```
<term name="bill">Gesetzentwurf</term>
<term name="legislation">Gesetz</term>
<term name="regulation">Verordnung</term>
```

The 'series-creator' is really tricky. AFAIK, there is no name for it in German. We mostly call it the "Showrunner" or "Produzent". So I am very unsure and undecided about
```

<term name="series-creator">
      <single>series creator</single>
      <multiple>series creators</multiple>
</term>

<term name="series-creator" form="short">
      <single>Ersch.</single>
      <multiple>Ersch.</multiple>
</term>	
```

... and so on.

I hope this helps.

Thanks for all the work you continue to put into it!

tillheilmann:
Great, thank you!

BTW: It just occured to me that the correct German word for 'thesis' is probably this:

`<term name="thesis">Qualifikationsarbeit</term>`

adam3smith:
I doubt we'll see much usage of the term either way, but I agree that Qualifikationsarbeit isn't widely used. Abschlussarbeit is probably best as a translation.

We just go with Dissertation in the German Translation of Zotero because there is no widely used term that covers all theses and doctoral dissertations are going to be by far the most common cited items, so that'd be the alternative here. That way, styles could do something like print the term when `genre` is empty and otherwise use genre. That's obviously not the most accurate translation, but might be the most useful for localized styles.

adam3smith:
For `series-creator`, I have seen "Schöpfer" (in the German Wikipedia entry for Aaron Sorkin), but that does sound odd to me. All other options aren't accurate (both Showrunner and Produzent can differ from the original creator of a series)

Issue #317, opened 2024-07-08
Add `editor-tranlsator` terms


bwiernik:
CSL 1.0.2 added `editor-translator` as a defined creator type. (Processors should update to automatically populate this variable when `editor` and `translator` are identical.)

Currently, we have the term `editortranslator` in locale files. We should copy these lines and add `editor-translator`

bwiernik:
looks good

PR #367, opened 2025-09-04
Update locales-af-ZA.xml

POBrien333:
Adds new `editor-translator` term to all locales. Via #317

### DIFF
--- a/locales-af-ZA.xml
+++ b/locales-af-ZA.xml
@@ -560,6 +560,10 @@
       <single>redakteur &amp; vertaler</single>
       <multiple>redakteurs &amp; vertalers</multiple>
     </term>
+      <term name="editor-translator">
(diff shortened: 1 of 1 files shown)


bwiernik:
Looks good. Can you replicate for other locales?

bwiernik:
If the short or verb-short forms are missing, they don't _need_ to be added but I don't see any harm in doing so

POBrien333:
Gotcha. I've added it to two locales. The Pakistani ones I'm not sure about.

bwiernik:
There are no short forms at all in that locale so it's fine

POBrien333:
Hi
As you'll have noticed, I am asking more people to add translations to the locales when they make PRs for citation styles. Any chance we could get this PR merged in order to avoid any merge conflicts with later PRs and extra work to solve them then?

PR #391, opened 2025-10-16
Revert addition of sr Latin and Cyrillic scripts

bwiernik:
@nikola-grubor The current testing pipeline isn't compatible with specifying a script. We will need to update the pipeline before we can restore these. 

@retorquere This is caused by the `to_s` method for `CSL:Locale` from csl-ruby. When the locale is normalized, it splits the locale name on `-` and keeps only the first 2 elements. Do you see a workaround?

@inukshuk Would it be possible to update csl-ruby to accommodate locales that match: `/([a-zA-Z]{2,3})-?(?=[a-zA-Z]*)/`


inukshuk:
Will we have a `sr-RS` locale file? Which file should we select if the language is specified as `sr` for example?

bwiernik:
Yes, we have a few locales that don't have a 2 letter language code, so supporting 3 letter language codes is needed

inukshuk:
OK that's done in 2.2.1

bwiernik:
Thanks @inukshuk! That worked great

Issue #352, opened 2025-07-07
Synchronize locales with en-US


adunning:
As @bwiernik flags in #301, it would be fairly straightforward to expand <https://github.com/citation-style-language/locales/blob/9d5a7903299c1c56598bf758a3153358a57908e2/util/add-locale-terms.py> to restructure all the locales on the basis of the `en-US` file.

I am raising this issue to ensure this doesn't fall through the cracks. I should be able to get to this eventually, but anyone else is welcome to take this on.

bwiernik:
@zepinglee go ahead!

PR #394, opened 2025-10-16
Synchronize locales with en-US

zepinglee:
As requested in <https://github.com/citation-style-language/locales/pull/301#issuecomment-3017157714> and <https://github.com/citation-style-language/locales/issues/352>, this PR sorts `<term>` elements in all locale files according to canonical order from `locales-en-US.xml`.

- Semantic preservation: only the order of terms is changed and the content of each term remains the same as before. This is checked via the [`get_terms_dict()`](https://github.com/zepinglee/csl-locales/blob/98c6231300ea35da92bd33574aab6e772c7948f1/util/sort_terms.py#L236C5-L236C19) fucntion.
- The comments  are also preserved.

### DIFF
--- a/locales-bal-PK.xml
+++ b/locales-bal-PK.xml
@@ -5,7 +5,7 @@
       <name>عثمان</name>
     </translator>
     <rights license="http://creativecommons.org/licenses/by-sa/3.0/">This work is licensed under a Creative Commons Attribution-ShareAlike 3.0 License</rights>
-    <updated>2025-10-16T03:12:52+00:00</updated>
+    <updated>2025-10-18T05:53:07+00:00</updated>
   </info>
   <style-options punctuation-in-quote="true"/>
   <date form="text">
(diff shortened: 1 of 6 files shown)


bwiernik:
Awesome! Thank you!

> The short form of et-al term in the en and fr locales does not exist in other locales. Are they necessary?

I think we can delete these short forms

> There are two "authors" terms which are usually missing in other locales. I'm not sure they are required in those languages.

Let's delete these

> The following terms are duplicate.

Good catch. Let's fix these in a separate PR

Could you update from master and run the script on the two new sr locales?

zepinglee:
> > The short form of et-al term in the en and fr locales does not exist in other locales. Are they necessary?
> 
> I think we can delete these short forms
> 
> > There are two "authors" terms which are usually missing in other locales. I'm not sure they are required in those languages.
> 
> Let's delete these
> 
> Could you update from master and run the script on the two new sr locales?

I've rebased from master branch and run the sorting again.

BTW, why not add an empty "author" term just like the "timestamp" term?

https://github.com/citation-style-language/locales/blob/10ce56fb7730fd3524548d37386c23a792afe621/locales-en-US.xml#L370-L371

bwiernik:
Okay can go ahead and add that please and include a "generally blank" comment like timestamp?

PR #403, opened 2025-12-16
Update locales-lv-LV.xml

ksedleni:
Changed months from locative to nominative

### Description

Briefly describe the changes you're proposing.

### Checklist

- [ ] Check that you've updated the date under `<updated>` in the `<info>` block. Be sure to preserve the original date-time formatting (for example, `2025-09-22T22:06:38+00:00`)
- [ ] If possible, please include references to dictionaries, style guides, or similar that inform your translations (see the en_US locale for examples). This helps us to keep locales up to date over time.

### DIFF
--- a/locales-lv-LV.xml
+++ b/locales-lv-LV.xml
@@ -726,18 +726,18 @@
     <term name="translator" form="verb-short">tulk.</term>
 
     <!-- LONG MONTH FORMS -->
-    <term name="month-01">janvārī</term>
-    <term name="month-02">februārī</term>
-    <term name="month-03">martā</term>
-    <term name="month-04">aprīlī</term>
-    <term name="month-05">maijā</term>
-    <term name="month-06">jūnijā</term>
-    <term name="month-07">jūlijā</term>
-    <term name="month-08">augustā</term>
-    <term name="month-09">septembrī</term>
-    <term name="month-10">oktobrī</term>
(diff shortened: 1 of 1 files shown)


ksedleni:
These are used in indicating publication date. One would use locative (2025. gada 15. decembrī) when describing an event that took place on a paticular day or when documenting when something happened (for instance in letters date is usually in this format, indicating that the letter was written on this day). However, in all other situations usually nominative is used, indicating not the placement of an event in time but using the date as an attribute. Thus, when writing bibliographies all dates are assumed to be in nominative - they are attributes of the texts. Therefore, if they are used in bibliographies, it must be in nominative: 2025. gada 15. decembris, 2025. gada 1. maijs etc. English by the way uses the same principle: the date is used without a preposition in bibliographies (May 20, 2025 rather than on May 20, 2025).

ksedleni:
1. yes, in all these cases nominative may be used. 
2. In this case if the day is mentioned as a part of a narrative, it should be in locative (e.g. "last checked May 25, 2025" would be "pēdējo reizi skatīts 2025. gada 25. maijā". or your case: "Retrieved 6 May 2025 from https://example.com/" would probably be "lejupielādēts 2025. gada 6. maijā".

bwiernik:
Thanks, I'll note that this is an area for potential future improvement in CSL

PR #404, opened 2025-12-19
Et ee available at translation

aluhamaa:
### Description

Changed one field in translation

### Checklist

- [X] Check that you've updated the date under `<updated>` in the `<info>` block. Be sure to preserve the original date-time formatting (for example, `2025-09-22T22:06:38+00:00`)
- [ ] If possible, please include references to dictionaries, style guides, or similar that inform your translations (see the en_US locale for examples). This helps us to keep locales up to date over time.


aluhamaa:
> Thanks! Do you think you could also provide some of the other missing translations for this locale?

Yes, I think it makes sense to try. I just need to consult with a few people to ensure I have things correctly.

PR #415, opened 2026-01-25
Update locales-de-CH.xml

matthiasmoier:


adam3smith:
I think most of these are improvements in any language; I'm skeptical of "Rezension" -- while Review isn't super common in German, Rezension is more specific and may not fit the bill.
Frühling vs. Frühjahr is locale specific; it should remain Frühjahr in de-DE

PR #427, opened 2026-05-05
patch-1

sergioanonlijo:
I have added translations of some terms and standardized some abbreviations and the use of particles. I consulted the RAE, RAG, and SciELO.


POBrien333:
continued in #435

PR #436, opened 2026-06-01
Update locales-sv-SE.xml

neonfabrik:
### Description

- Translation of missing general terms, item types, locators, variables, and roles.

### Checklist

- [x] Check that you've updated the date under `<updated>` in the `<info>` block. Be sure to preserve the original date-time formatting (for example, `2025-09-22T22:06:38+00:00`)
- [x] _(Not great but better)_: If possible, please include references to dictionaries, style guides, or similar that inform your translations (see the en_US locale for examples). This helps us to keep locales up to date over time.


POBrien333:
thanks!

